PR #3, opened 2018-02-06
Not overwriting project

henpe071:
When creating a new project, before the prompt accepts the answer, it checks if the path already exists. If it does, it alerts the user and asks if he/she wanna continue and overwrite the old one, open the old one or just cancel the action.

### DIFF
--- a/ViAn/GUI/projectdialog.cpp
+++ b/ViAn/GUI/projectdialog.cpp
@@ -66,16 +66,26 @@ void ProjectDialog::ok_btn_clicked() {
         QMessageBox msg_box;
         msg_box.setModal(true);
         msg_box.setText("This folder already exist. Are you sure you wan't to continue?");
-        msg_box.setInformativeText("This will overrite the current project.");
-        msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
+        msg_box.setInformativeText("Yes will overrite and Open will open the current project.");
+        msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Open);
         msg_box.setDefaultButton(QMessageBox::No);
         int reply = msg_box.exec();
+        if (reply == QMessageBox::Open) {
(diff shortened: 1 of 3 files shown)


ViAn/GUI/projectdialog.cpp
```diff
@@ -57,9 +59,33 @@ void ProjectDialog::browse_btn_clicked() {
 }
 
 void ProjectDialog::ok_btn_clicked() {
+    QString m_dir = path_text->text() + "/" + name_text->text() + "/";
+    QDir pathDir(m_dir);
+    if (pathDir.exists()) {
+        // Create confirmation dialog since the path already exists
+        QMessageBox msg_box;
+        msg_box.setModal(true);
+        msg_box.setText("This folder already exist. Are you sure you wan't to continue?");
```
fredrikiselius: want

ViAn/GUI/projectdialog.cpp
```diff
@@ -57,9 +59,33 @@ void ProjectDialog::browse_btn_clicked() {
...
+        msg_box.setDefaultButton(QMessageBox::No);
+        int reply = msg_box.exec();
+        // Open the already existing project
+        if (reply == QMessageBox::Open) {
+            qDebug() << path_text->text();
+            if (path_text->text() == "") {
+                m_dir = "C:" + m_dir;
+            }
+            emit open_project(m_dir + name_text->text());
+            qDebug() << m_dir;
```
fredrikiselius: Remove debug messages

fredrikiselius (commented):
When a project is overwritten the treestate from the old project is still saved. It needs to be removed else placeholder items will appear in the project tree when loading.

henpe071:
I fixed the comment and removed the debug messages. I will wait with the problem that the project still partly remains after it's been overwritten. That problem should be easily fixed once we get the temporary save files. So instead of doing a work-around here to fix it I'll just wait.

PR #5, opened 2018-02-06
Recent project window

henpe071:
Made the recent projects-window a little more intuitive by adding a new button. This new button will open the selected project. you can still double click to open it but now there's also a button. Aside from that I also disabled most of the actions in the menus when there's not a project loaded. This because, firstly, you didn't need those options since you couldn't use them and secondly since it's now much easier to understand that you have to create or open a project to do anything since nothing else works.


fredrikiselius:
Different views and actions should not be disabled, instead, a user should be forced to work in a project. The program should never be able to enter a state where a project doesn't exist. 

The "open project" button in the recent project window should be changed to "browse"

PR #9, opened 2018-02-15
Queue widget is now a dockable

henpe071:
Queue widget is now a dockable window but otherwise works the same. 

Remove the abort analysis button when no analysis are loaded.
Changed draw-analysis-area cursor to a crosshair.
Made the button toggable and added a "are you sure" pop up after analysis area been drawn.


fredrikiselius:
* Should the queue widget always close when the queue is empty?
* It seems that the draw square is kept on the video when adding a second analysis to the queue.

PR #10, opened 2018-02-20
Save analysis area

henpe071:
If an analysis is done within an interval or a bounding box , the interval and box are saved and can be showed by turning them on in the contex menu of the analysis. 

When an analysis is done it will check if an analysis with that name already exist in the target folder and if so add a number to the end so it doesn't overwrite.

Fixed so the analysis is done on the correct area when video is zoomed in.

Analysis interval, the one that shows where the analysis is done, is now a seperate interval. Intervals are now saved as pairs instead of analysisInterval.


ViAn/Analysis/analysismethod.h
```diff
@@ -35,7 +35,8 @@ class AnalysisMethod : public QObject ,public QRunnable{
     bool paused = false;
 
 protected:
-    AnalysisInterval interval;
+    std::pair<int, int> interval;
```
fredrikiselius, line 38: Why not use the AnalysisInterval? It contains the same information

ViAn/GUI/Analysis/analysisslider.cpp
```diff
@@ -52,20 +52,39 @@ void AnalysisSlider::paintEvent(QPaintEvent *ev) {
         }
     }
 
+    // Draws the analysis interval
```
fredrikiselius, line 55: Is this the same code as for the interval drawing? If so, it should be reused

ViAn/GUI/Analysis/analysisslider.cpp
```diff
@@ -52,20 +52,39 @@ void AnalysisSlider::paintEvent(QPaintEvent *ev) {
...
+    if (show_ana_interval) {
+        brush = Qt::darkMagenta;
+        double first = (groove_rect.left()+(double)m_ana_interval.first*c);
+        double second = (groove_rect.left()+(double)m_ana_interval.second*c);
+        if (m_ana_interval.first != -1 && m_ana_interval.second != -1 && m_ana_interval.first <= m_ana_interval.second) {
```
fredrikiselius: It quite hard to read the code with all these conditions that doesn't really say anything. 

 m_ana_interval.first != -1 && m_ana_interval.second != -1

Could for example be refactored into a "valid_interval" function or similar.

ViAn/Analysis/analysismethod.h
henpe071: I felt like i didn't need/want the extra things in the AnalysisInterval. The interval also isn't an AnalysisInterval in the way we used it before. And lastly, before the interval was sometimes a pair and sometimes an AnalysisInterval which wasn't really necessary. Now it's the same type all the time.

Both works, I just chose a pair.

henpe071:
Structured the code a little better in the analysisslider and added the show/hide details in the view menu. The show/hide details now also remains even when changing video or analysis.

henpe071:
TODO fix crash

PR #11, opened 2018-02-20
Fixed the brightness/contrast window

henpe071:
Fixed the dialog with auto updating sliders and made it remember old changes.


fredrikiselius:
Looks good as it is.

PR #14, opened 2018-02-27
Drawing tool

henpe071:
Made changes and added new functionality to the drawing tool.

- Added a new hand tool. This works like an edit tool. While active click a drawing to hightlight it as active drawing. The active drawing will have its color inverted as an indicator. 

The hand tool have 3 major functionalities. 
- By left-clicking a drawing and moving the mouse you can move it. 
- By right-clicking and moving the mouse you can edit the size of the current drawing.
- By using the scroll wheel you can change the thickness of the lines on the current drawing. If the thickness goes to negative the shape is filled (only rect and circle).

- The clear all drawings in the toolbar now instead deletes the current drawing. The clear all action is still available in the menu.
- Connected the menu options with their respective action in the toolbar. Now the action in the toolbar is selected when something is selected in the menu. 

----- Things not yet added -----
I'm not exactly sure how to do these things.

- Edit the text in the text drawing.
- Resize text drawings.
- Resize pen drawings.
- Uncheck the action in the toolbar.
- Show the current color in the toolbar.
- Add the delete current action to the undo/redo list. 
- (Maybe) Add movement and resizing the the undo/redo list.


fredrikiselius:
Nothing to add really. Looks fine

PR #16, opened 2018-03-01
Prompt save

### LINKED ISSUE #13: Renaming items in the project tree affects GUI only
When renaming items in the project tree the change affects GUI only. As such the name change is not persistent and will changes back when the project is reloaded.

fredrikiselius:
Added a save prompt that will appear when the user is trying to close a project that has unsaved changes. Also fixes #13.


ViAn/Project/bookmark.cpp
```diff
@@ -68,18 +68,24 @@ std::string Bookmark::get_description() {
  * @param text
  */
 void Bookmark::set_description(const std::string& text) {
+    qDebug() << "NEW BOOKMARK DESCRIPTION " << QString::fromStdString(text);
```
henpe071: remove debug message

henpe071 (commented):
There needs to be one more repaint when opening a new project and the save prompt that you get when trying to create a new project is underneath the new project window. These things should probably however fix themselves when you merge.

Another thing I noticed is that the program crashes when you save it or close it while an analysis is running. This does however exist even in development. I just thought that since you were doing checks it might be appropriate to do aswell otherwise we'll just put it up as a bug issue.

Apart from these it looks fine.

ViAn/Project/bookmark.cpp
fredrikiselius: The bug you mentioned, whilst important to fix, should be solved in another branch. Will add an issue.

PR #22, opened 2018-03-14
Tag from drawing

henpe071:
Right click a video and in the context menu there is an option to add all current drawings to a tag. This tag works like a normal tag except it can't be changed manually. Currently the tags does not update automatically as new drawings are drawn or old ones removed. Instead there is an option in the context menu for the drawing tag to update it. 
The drawing tags can be renamed and saved just like normal tags.

The save functions are a little bit cleaned up. 
The save file now saves the correct path instead of the temporary path's.
SAVE_TYPE is removed and ANALYSIS_TYPE is used instead. Didn't need two enum type specifying an item.
No longer saves the frames where there are no drawings.

Also fixed a small thing with add folder.

TODO
Undo/Redo: since the undo/redo functionality gets removed in the Drawing widget branch it is not fixed here. It will not work perfectly with the drawing tags.
Remove treeitems: Removing tags, drawing tags and analyses from the treewidget is not yet implemented. That is a functionality that will get its own branch since it needs a little more testing and isn't fully related to this feature.
Export as bookmarks:


ViAn/Analysis/analysismethod.cpp
```diff
@@ -2,14 +2,16 @@
...
-AnalysisMethod::AnalysisMethod(const std::string &video_path, const std::string& save_path)
+AnalysisMethod::AnalysisMethod(const std::string &video_path, const std::string& tmp_save_path, const std::string& save_path)
 {
     m_source_file = video_path;
     std::size_t index = video_path.find_last_of('/') + 1;
     std::string vid_name = video_path.substr(index);
     index = vid_name.find_last_of('.');
     vid_name = vid_name.substr(0,index);
-    m_save_path = save_path+vid_name +"-motion_analysis";
+    m_ana_name = vid_name + "-motion_analysis";
```
fredrikiselius: "-motion_analysis" should be a constant that is appended depending on the type of analysis

ViAn/Analysis/analysismethod.cpp
```diff
@@ -159,16 +162,17 @@ void AnalysisMethod::run() {
  * Checks if path exists and if it does adds a number to the end to prevent
  * that path overwrites the old one
  */
-std::string AnalysisMethod::check_save_path(std::string path, int increment) {
-    std::string new_path = path + std::to_string(increment);
+std::string AnalysisMethod::check_save_path(std::string name, int increment) {
```
fredrikiselius: Isn't this function already implemented and used by bookmarks? Anyway, this function should probably not be part of the AnalysisMethod class since it dosn't actually do anything with an analysis. Could be added to misc instead?

ViAn/GUI/projectdialog.cpp
```diff
@@ -32,6 +32,8 @@ ProjectDialog::ProjectDialog(QWidget *parent) : QDialog(parent) {
     btn_box->addButton(QDialogButtonBox::Ok);
     btn_box->addButton(QDialogButtonBox::Cancel);
 
+    path_text->setText("C:/");
```
fredrikiselius, line 40: Should be os default documents folder

ViAn/GUI/projectwidget.cpp
```diff
@@ -585,6 +617,44 @@ void ProjectWidget::context_menu(const QPoint &point) {
...
+        }
+    }
+    add_basic_analysis(vid_proj, basic_tag);
+}
+
+/**
+ * @brief ProjectWidget::update_tag_drawing
+ * Update the drawing tag so it matches with all the current drawings
+ */
+void ProjectWidget::update_tag_drawing() {
```
fredrikiselius: Almost identical with `ProjectWidget::tag_drawings()`, could be merged into one function with a check if the tag should be updated?

ViAn/Project/videoproject.cpp
```diff
@@ -126,14 +126,17 @@ void VideoProject::read(const QJsonObject& json){
...
         BasicAnalysis* analysis;
-        SAVE_TYPE save_type = (SAVE_TYPE)json_analysis["save_type"].toInt();
+        ANALYSIS_TYPE save_type = (ANALYSIS_TYPE)json_analysis["analysis_type"].toInt();
         switch(save_type){
-        case INTERVAL:
+        case TAG:
             analysis = new Tag();
             break;
-        case DETECTION:
+        case MOTION_DETECTION:
```
fredrikiselius, line 134: I think `DETECTION` was used here since it would make it simpler if more detection types are added

fredrikiselius (commented):
Looks fine, some small issues mentioned below. Should a new drawing tag be created each time "tag drawings" is used?

ViAn/Analysis/analysismethod.cpp
henpe071: Is there a better way to do this? Beacuse this function is not the best. Also how does what you're working on affect this since the temporary map will be removed?

ViAn/GUI/projectdialog.cpp
henpe071: How do you set this to os default? Just writing "C:/..../documents/" will not always work.

ViAn/GUI/projectwidget.cpp
henpe071: These aren't as identical as you say. The only thing identical is the loop but it might still be able to simplify.

henpe071:
Fixed the comments.

fredrikiselius:
Looks fine

PR #25, opened 2018-03-19
Always in project

### LINKED ISSUE #15: Always in a project
When starting ViAn an empty project should be created. This way there will always be a project even if the user doesn't creates a new or opens an old project from the Recentproject window.

henpe071:
A default project is always created when the program starts. This means that there no longer is a need to first create a project before start working. The "not saved" - prompt will still ask you to save the project before closing it and the save prompt will ask for a name and path before saving. 

Edit: fixes #15

### DIFF
--- a/ViAn/Project/Test/projecttestsuite.cpp
+++ b/ViAn/Project/Test/projecttestsuite.cpp
@@ -21,6 +21,8 @@ void ProjectTestsuite::add_remove_vid_proj_test(){
     proj->remove_video_project(vp3);
 
     QCOMPARE(proj->m_videos.size() , unsigned(0));
+
+    delete proj;
 }
 
 void ProjectTestsuite::add_remove_report_test(){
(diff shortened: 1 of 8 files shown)


ViAn/Project/Test/projecttestsuite.cpp
```diff
@@ -41,10 +43,14 @@ void ProjectTestsuite::add_remove_report_test(){
     proj->remove_report(idx3);
 
     QCOMPARE(proj->m_reports.size(), unsigned(0));
+
+    delete proj;
 }
 
 
 void ProjectTestsuite::save_load_test(){
+    // TODO Fix
```
henpe071, line 52: I looked a little at theses test cases and they all created a temporary dir but never removed it which eventually gave me error because i reach the max limit. The added deletes removed the temp folder except in this one. I know what the problem is and the solution lies in the tag-drawing branch. I'll fix it after these all are merged and just added a reminder.

fredrikiselius:
Nothing to comment on really, looks good

PR #26, opened 2018-03-19
Connected all actions related to mouse tools

henpe071:
- Moved some actions from the button layout up to the toolbar.
- Added a panning tool.
- All actions in the toolbar are connected so only one can be active at a time.
- Added a No tool action.
- Added diferent cursors for the different actions.
- Added a zoom out tool to zoom out multiple times.
- Using the scroll wheel to changes the thickness now only works while the edit tool is active. Still the current drawing that gets changed.
- Added the ability to change the zoom level with the scroll wheel. This works while zoom in, zoom out or the panning tool are active.


fredrikiselius:
Looks good

PR #27, opened 2018-03-20
Multiple video widgets

henpe071:
NOT READY TO MERGE!
It's possible to open multiple seperate videowidgets. To do so, rightclick the video you wanna open and choose "open video in widget" and a new widget will appear. This is not limited to only one new video widget. 

The new video widget have the same functionality as the normal one. The video buttons, slider zoom and all that works. It's also possible to create a tag, a bookmark and export frames. 
Currently it's not possible to tag frames, do an analysis or draw drawings.

I mostly put this one up for review since I'm not sure if there's a better way to handle the threads. If it is ok I will clean the branch and make it ready for merging. 
Also, despite the branch name the new videowidgets are not dockable since it was requested that way.


fredrikiselius:
* Remove all buttons that can't be used in the dockable video widget
* Limit the number of widgets that can be opened?
* Catch and handle OpenCV error that occurs when out of memory due to many widgets: `OpenCV Error: Insufficient memory (Failed to allocate 24883200 bytes) in OutOfMemoryError`
* Should the user be able to bookmark from every widget? If so it must later be added that a clicked bookmark causes a jump to the frame in the correct video widget.
* There are some problems that occur when the program is closed whilst multiple videos are playing. This must be handled correctly

henpe071:
Yes, on most of it.
What upper limit of widgets should I set? Since this goes hand in hand with the out of memory issue the amout it can handle varies.
Regarding the bookmarks, here we need to decide what actions a user should be able to use in the extra widgets and how they should work. My suggerstion is that you might be able to bookmark from the floating widgets but if you jump from the bookmark the main videowidget changes. (and this is the easiest after removing the funcionality completely). The same thing applies to tags and analyses.

fredrikiselius:
Perhaps limiting the number of widgets is a bad idea. A preferable solution would be to check for the required memory for opening a new video and then limiting based on that. I think we should hold off on merging until we've fixed that.

ViAn/GUI/mainwindow.cpp
```diff
@@ -212,9 +214,7 @@ MainWindow::~MainWindow() {
     delete project_wgt;
     delete analysis_wgt;
     delete bookmark_wgt;
-    delete status_bar;
```
henpe071, line 214: Is it better  to delete object manually like this or let the program clean them up by itself? They are still deleted as the code is now.

fredrikiselius:
I've tested the new version and it still crashes when I open enough new widgets. I get the following messages:

`QThread::start: Failed to create thread (The access code is invalid.)`
`createDIB: CreateDIBSection failed.`

A crash also occurs if 3 or more widgets are playing and the main windows is closed (Without any msgs).

It also crashed when I closed a single widget whilst a video was playing in it. I haven't been able to replicate that though.

Finally when there is insufficient memory the widget still appears but without the video. Is this the desired behaviour or should it simply close? (This is when the "Failed to copy new frame" print appears.)

henpe071:
I hope i didnt break it. didnt wanna merge with development

PR #28, opened 2018-03-26
Drawing widget

henpe071:
Added a widget that shows all drawings on the current video project. The drawing are store so all drawings on a frame are collected under their frame number which is a parent in the tree.

-The widget is dockable.
-The frame numbers are sorted.
-Clicking a frame number in the tree will jump to that frame.
-Clicking a drawing will jump to the correct frame and hightlight the drawing with a dashed line rectangle.
-Right clicking an item will open a context menu where you can choose rename or remove.
-Removing the frame number will delete all children and deleting the last drawing under a frame will delete the frame as well.
-Renaming the text item will change the text on the overlay.
-Clicking an item will automaticaly hightlight the hand tool
-With the hand tool you can move drawings with left click, edit with right click and change the line thickness with the scroll wheel.
-The text will change size with the scroll wheel.
-Current item can only be selected from the widget.
-Removed the undo/redo functionality


ViAn/GUI/drawingwidget.cpp
```diff
@@ -0,0 +1,361 @@
...
+    connect(m_overlay, SIGNAL(new_drawing(Shapes*, int)), this, SLOT(add_drawing(Shapes*, int)));
+}
+
+void DrawingWidget::clear_overlay() {
+    if (m_overlay != nullptr) {
+        save_item_data();
+        disconnect(m_overlay, SIGNAL(new_drawing(Shapes*, int)), this, SLOT(add_drawing(Shapes*, int)));
+        m_overlay = nullptr;
+    }
+    clear();
```
fredrikiselius, line 37: The clear function only removes gui items, not underlying data. Haven't check whether this is done properly or not yet. If not it should be

If that is the intended use ofc

ViAn/GUI/drawingwidget.cpp
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+/**
+ * @brief DrawingWidget::add_drawings_to_frame
+ * Adds all drawings to the drawing widget tree
+ * @param f_item
+ */
+void DrawingWidget::add_drawings_to_frame(FrameItem* f_item) {
+    std::vector<Shapes*> list = m_overlay->get_overlays()[f_item->get_frame()];
```
fredrikiselius: This type of indexing `m_overlay->get_overlays()[f_item->get_frame()]` is dangerous. If the frame number does not exist in the overlay map, the program will probably throw an exception and crash. 

Not entierly sure though, but its worth checking

ViAn/GUI/drawingwidget.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+/**
+ * @brief DrawingWidget::item_changed
+ * @param item
+ * Slot function called on the signal itemChanged.
+ * Updates the item with the newly entered name
+ */
+void DrawingWidget::item_changed(QTreeWidgetItem* item) {
+    auto a_item = dynamic_cast<ShapeItem*>(item);
+    a_item->rename();
```
fredrikiselius: This is quite confusing as it is easy to assume that this will rename the actual tree item. But instead it actually renames the data item. Change the name of the function to increase readability. This goes for all drawing items

ViAn/GUI/drawingwidget.cpp
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+/**
+ * @brief DrawingWidget::dropEvent
+ * Handles drop events.
+ * Calls upon standard dropEvent for TreeItems.
+ * @param event
+ */
+void DrawingWidget::dropEvent(QDropEvent *event) {
```
fredrikiselius: Remove?

ViAn/GUI/mainwindow.cpp
```diff
@@ -519,12 +537,24 @@ void MainWindow::redo() {
...
+}
+
+void MainWindow::delete_drawing(Shapes *shape) {
+    video_wgt->set_delete_drawing(shape);
 }
 
-void MainWindow::delete_drawing() {
-    video_wgt->set_delete_drawing();
+void MainWindow::delete_current_drawing() {
+    video_wgt->set_delete_drawing(nullptr);
```
fredrikiselius: ??

fredrikiselius (commented):
* If a selectable drawing is marked, it is not deselected when a pen drawing is marked
* Undo/redo has no functionality?
* It is possible to step between drawings in the drawing widget using the key arrows. The selection is thus changed but the marked drawing in the video view is constant. It should changes in accordance with the widget.
* Drawings are created on the pressed event, thus making it possible to create non-existing drawings. Should probably be created on the release event instead.
* The dashed rectangle should have a black non-dashed rectangle underneath to create better contrast
* Would be useful if the drawing widget scrolled to the `FrameItem ` with the current frame number. It might also be good if all other items are closed and the current expanded. It gets rather messy when many frames have drawings on them.
* Moving drawings does not trigger the save prompt if the program is closed (neither does renaming the item in the drawing widget list)
* Not possible to deselect a drawing from the drawing widget

More comments below:

ViAn/GUI/mainwindow.cpp
henpe071: they both call the same function which does a check (shape != nullptr) to know if it should delete current or the shape. And since mainwindow where delete_current_drawing is from don't know the current drawing i kinda had to do something like this somewhere. Though this might not be the best way.

ViAn/GUI/drawingwidget.cpp
henpe071: I know and this is what i looked at earlier this week in the "remove-items" branch. Right now if I put a delete in the shapeItem's destructor deleting the shape it will complain. But it's not perfect and there are still some mem leaks somewhere.

henpe071 (commented):
--The pen have a small bug that should be fixed but my thought was that the pen drawings would not be selected like the others, since the pen drawings doesn't have the start/end point the same way as the others have. And therefore the dashed rectangle would not capsule the drawing correctly.
--Undo/redo functionality is removed since it doesnt work with the widget. The buttons are still there however but they should be removed.
--The keyboard shortcuts is one thing i forgot to do but I should do that in a new branch
--What do you mean with non-existing drawings? Yes they will be pixel-big if you just click but they should be easy to edit or remove from the widget.
--Yes
--Yes
--Will look into.
--Can probably change to if you click in the widget but not on an item it will deselect the current

ViAn/GUI/drawingwidget.cpp
henpe071: The data are eventually deleted so it's not all leaked but it's not done optimally. the fix should be in the "remove" pull request which aims to fix a lot of memory leaks and removings items and data.

henpe071:
Fixed some comments and made the pen drawings selectable and dragable. To do this i changed so they are save as points instead of pairs of points.

PR #32, opened 2018-04-10
Color indicator

henpe071:
Added another column with a small color indicator for the current color of the shape. By clicking the indicator the color dialog pops up and the user will be able to change the color of the drawing.
Also added small icon for each drawing indicating which type they are.

The branch is a continuation on the drawing widget pull request.


fredrikiselius (commented):
For what I can see all different item classes are subsclasses of ShapeItem. As such common function as update_shape_color and the qpixmap member should be part of the ShapeItem class instead.

PR #35, opened 2018-04-13
Project file handling

fredrikiselius:
* Creating a new project will no longer prompt the user for a save path and project name. This is done at save instead.
* Files for new projects are now stored inside temporary folders. These folders and files are deleted when the user saves or closes the project.

### DIFF
--- a/ViAn/GUI/projectwidget.cpp
+++ b/ViAn/GUI/projectwidget.cpp
@@ -45,7 +45,7 @@ ProjectWidget::~ProjectWidget() {
 
 /**
  * @brief ProjectWidget::new_project
- * Creates a create project dialog
+ * Creates a new empty project if current project is closed/user closes it
  */
 void ProjectWidget::new_project() {
 //    ProjectDialog* proj_dialog = new ProjectDialog();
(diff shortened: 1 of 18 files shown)


ViAn/GUI/projectwidget.cpp
```diff
@@ -459,6 +440,17 @@ bool ProjectWidget::prompt_save() {
     return ok;
 }
 
+bool ProjectWidget::prompt_close() {
```
henpe071: Use "ProjectWidget::message_box".

henpe071 (changes requested):
- I one time managed to get it to fail to save the analysis but I can't replicate it. 
- The save paths have an extra "/" in them. Nothing major but it's unnecessary.
- If the first thing you do after clicking new project is closing it, it will still prompt the user with the unsaved promt. Not needed.
- Just closing the recent project dialog will not open a temporary project. Only when the "new project" is clicked.
- If you try to save your temporary project with the same name as a project that alredy exists you get a prompt and the open button in that one doesn't work. It will just open a new temporary project and not close the current project. It should open the already existing project with that name.
- It is possible to save a project with an empty name.
- There needs to be another default value or case for when the temporary folder is not valid. Now if the temp folder is invalid when you close it the folder "" will be cleared. DO NOT try it. 
- Bookmarks created after a project have been saved from temp to "real" are still created in a temp folder, though a new one. This may be an issue for later when working more on the bookmarks.
- If, when saving the temp project, you add for ex "/1" at the end of the path the treestate will not be saved.

henpe071 (approved):

henpe071:
Looks good

PR #39, opened 2018-04-19
New zoom wheel

henpe071:
The zoom with the mouse wheel works as requested.

- Mouse wheel: Pan up & down.
- Mouse wheel + Shift: Pan left & right.
- Mouse wheel + control: Zoom in & out centered around the cursor.


fredrikiselius:
looks fine

PR #40, opened 2018-04-20
Frame num save

henpe071:
The frame number, brightness and contrast are now saved when switching between videos.
Also change videos with the arrow keys.


fredrikiselius:
Looks fine. When loading a video with a stored frame number the first frame is still displayed. Should probably be fixed later on. The playback slider appears to be jumping as well

fredrikiselius:
Caused major bugs after merge with development. Will revert development and restore this branch

PR #44, opened 2018-04-27
Remove, clean

henpe071:
Added the ability to remove analyses and tags, both versions. 
Fixed some code and cleaned it up a little.
Looked a lot at how different objects are created and removed and hopefully fixed the memory management a little by adding more destructors which cleans up the correct thing.


ViAn/GUI/mainwindow.cpp
```diff
@@ -177,6 +176,8 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent){
 
     // Recent projects menu
     rp_dialog = new RecentProjectDialog(this);
+    rp_dialog->setAttribute(Qt::WA_DeleteOnClose);
+    connect(rp_dialog, &RecentProjectDialog::open_project, project_wgt, &ProjectWidget::open_project);
```
fredrikiselius: This row is a duplicate of line 186

ViAn/GUI/projectwidget.cpp
```diff
@@ -678,26 +688,71 @@ void ProjectWidget::remove_item() {
...
-//            else if (item->type() == TAG_ITEM || item->type() == ANALYSIS_ITEM) {
-//
-//            }
-            delete item;
+/**
+ * @brief ProjectWidget::remove_tree_item
+ * @param item
+ * Recursive function for removing item and all of its children
+ */
+void ProjectWidget::remove_tree_item(QTreeWidgetItem* item) {
```
fredrikiselius, line 701: Quite a large function, should probably be refactored into function for each item type. Should also use a switch instead of if

ViAn/GUI/projectwidget.h
```diff
@@ -27,10 +27,7 @@ class VideoProject;
 class ProjectWidget : public QTreeWidget
 {
     Q_OBJECT
-    QTreeWidgetItem* clicked_item = nullptr;
-    QPoint* clicked_point = nullptr;
     QTreeWidgetItem* selection_parent = nullptr;
-    bool selecting = false;
```
fredrikiselius, line 33: Couldn't find any other changes using this variable. Unused?

ViAn/GUI/recentprojectdialog.cpp
```diff
@@ -40,6 +41,12 @@ RecentProjectDialog::RecentProjectDialog(QWidget* parent) : QDialog(parent) {
     connect(open_btn, &QPushButton::clicked, this, &RecentProjectDialog::on_open_btn_clicked);
 }
 
+RecentProjectDialog::~RecentProjectDialog() {
+    delete h_layout;
```
fredrikiselius, line 45: Each gui component will be deleted automatically by Qt if parented correctly. This isn't the case here but in future implementations we should set the parent correctly so we won't have to worry about cleanup

ViAn/Project/project.cpp
```diff
@@ -57,6 +65,7 @@ ID Project::add_video_project(VideoProject *vid_proj){
 void Project::remove_video_project(VideoProject* vid_proj){
     auto it = std::find(m_videos.begin(), m_videos.end(), vid_proj);
     if (it == m_videos.end()) return;
+    delete *it;
```
fredrikiselius, line 67: Do we always want to delete the video project here. Not just remove it from the project?

ViAn/Analysis/analysismethod.cpp
```diff
@@ -131,7 +131,6 @@ void AnalysisMethod::run() {
     if(*aborted){
         capture.release();
         emit analysis_aborted();
-        emit finito();
```
fredrikiselius, line 134: Was this signal not connected before? Can't find the removal of the connection line

ViAn/Project/Analysis/basicanalysis.cpp
```diff
@@ -14,9 +14,13 @@ BasicAnalysis::BasicAnalysis(const BasicAnalysis &other) :
     use_bounding_box(other.use_bounding_box){
 }
 
+BasicAnalysis::~BasicAnalysis() {
+    qDebug() << "basic analysis delete";
```
fredrikiselius: remove

ViAn/Project/project.cpp
```diff
@@ -7,8 +7,8 @@
  */
 Project::Project(){}
 
-//TODO Fix all these function names
 Project* Project::fromFile(const std::string &full_path){
+    qDebug() << "full path" << QString::fromStdString(full_path);
```
fredrikiselius: remove

fredrikiselius (commented):
Some minor comments and questions based on the code changes. 

When testing and switching between analyses the destructor is called on each switch. I'm not quite clear on the structure for the analyses so I don't know if this is wanted or not.

ViAn/GUI/projectwidget.h
henpe071: Yes

ViAn/Project/project.cpp
henpe071: I think so. This function is only used from the test cases though so I haven't looked at it that much.

ViAn/Analysis/analysismethod.cpp
henpe071: Was not used

henpe071:
When clicking an analysis the full analysis is loaded because the framewidget needs it to draw the detection boxes. So with each click a new analysis is created which have to be removed. So yes, that is the wanted behavior at least as the code works now.

PR #45, opened 2018-05-02
Hide drawings

henpe071:
Added a new column with a show/hide icon in the drawing widget. With this icon you can toggle the show/hide state of the drawing. Hidden drawings will not be visible and can't be moved.


fredrikiselius (approved):
Looks fine

PR #46, opened 2018-05-02
Added modifiers to drawing drawings

henpe071:
Shift modifier will make sure the drawing is symmetrical. Rects and circles will have equally long sides and lines and arrows will be drawn diagonally.
The control modifier only works for arrow and line. It will make them drawn either horizontally or vetically as a straight line.

### DIFF
--- a/ViAn/GUI/projectwidget.cpp
+++ b/ViAn/GUI/projectwidget.cpp
@@ -464,7 +464,7 @@ bool ProjectWidget::prompt_save() {
  */
 void ProjectWidget::tree_item_clicked(QTreeWidgetItem* item, const int& col) {
     Q_UNUSED(col)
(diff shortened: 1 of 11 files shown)


ViAn/GUI/projectwidget.cpp
```diff
@@ -464,7 +464,7 @@ bool ProjectWidget::prompt_save() {
  */
 void ProjectWidget::tree_item_clicked(QTreeWidgetItem* item, const int& col) {
     Q_UNUSED(col)
-    get_index_path(item); //Remove?
+    //get_index_path(item); //Remove?
```
fredrikiselius: Remove it instead

fredrikiselius (approved):
Looks fine, just remove the get_index_path

PR #47, opened 2018-05-02
Fix tools again

henpe071:
More changes to the drawing tool according to requests. 
- Merged the edit and select tool.
- Double clicking a drawing will always mark it as the current drawing change to edit tool and show the bounding box.
- Double clicking outside a drawing will change to the default tool (zoom).
- The bounding box is only visible when the edit tool is selected.
- Other small things around the bounding box and the current tool.

- Added copy paste drawings. 
- Copy drawings from one frame to another of one video to another.
- Ctrl + c / Ctrl + v.
- The pasted drawing will be on the same coordinates.


ViAn/GUI/Toolbars/drawingtoolbar.cpp
```diff
@@ -51,8 +51,8 @@ void DrawingToolbar::create_actions() {
     tools->addAction(rectangle_tool_act);
     tools->addAction(circle_tool_act);
     tools->addAction(line_tool_act);
-    tools->addAction(hand_tool_act);
-    tools->addAction(select_tool_act);
+    tools->addAction(edit_tool_act);
+    //tools->addAction(select_tool_act);
```
fredrikiselius, line 56: Remove or add todo if it should be used later

fredrikiselius (commented):
Looks fine, see comment below

ViAn/GUI/Toolbars/drawingtoolbar.cpp
henpe071: Added an issue to remove all the code associated with the old tools.

PR #49, opened 2018-05-21
Perform analysis dialog 2

henpe071:
Made updates to the analysis dialog dialog and the analysis menu.

- Changed the names of the analyses and moved the show/hide analysis details (and settings) to the analysis menu.
- Added the ROI analysis to the analysis menu (old quick analysis)
- Added the option to change all the settings for the analysis in the settings action. These settings are saved as the default settings and are used when starting a ROI analysis. 
- The dialog window have a reset button.
- Clicking the full analysis will open the analysis dialog. Here the user can customize the settings before starting an analysis.
- The current video is pre-selected.
- DETECT_SHADOWS is changed to a checkbox since it's a bool. (Also fixed to have a correct name)

- Analysis now hold an analysis_settings object with all of its settings. tex interval, boundingbox and sample frequency. This causes some change in how analyses are started and handled. 
- The context menu for an analysis have the option to toggle the analysis details which are global or...
- Show the analysis settings dockable widget which displays the settings for the current analysis.
- These two options are checkable.


ViAn/Analysis/analysismethod.cpp
```diff
@@ -3,44 +3,16 @@
...
     m_save_path = save_path;
-    add_setting("SAMPLE_FREQUENCY",1, "How often analysis will use frame from video");
-}
-
-/**
- * @brief AnalysisMethod::set_include_exclude_area
- * Sets an exlusion frame that will be used to exclude detections in a specific area of each frame.
- * @param points for the polygon that defines the exclusion area.
- */
-void AnalysisMethod::set_include_exclude_area(std::vector<cv::Point> points, bool exclude_polygon) {
```
henpe071, line 23: remains from how the old include/exclude worked

ViAn/Analysis/analysismethod.cpp
```diff
@@ -209,64 +177,10 @@ void AnalysisMethod::scale_frame() {
 /**
  *  Getters and setters
  */
-
```
henpe071, line 212: A lot of code are just moved from this file and motion detection to analysis settings.

fredrikiselius:
Looks good from what I can see. I have a question though. Is there only one instance of the settings object? And if so, is it possible to change the settings for an analysis at runtime? My initial thought is that we do not want that behaviour since it could generate confusing results.

henpe071:
For each analysis started a new analysis settings is created with the settings selected by the user or the current default. This means that it should not be possible to change an analysis' settings during runtime. Each analysis hold its own settings. The analysis settings widget will read this data and print it to the widget each time a new analysis is selected.

PR #50, opened 2018-05-23
Tags and state

henpe071:
Changed tags to display the frames it contains in the project widget as clickable items. This will make it easy to see which frames are tagged and untag single frames by removing them from the tag. 
Added a checkbox for the tags to see which tag is the current one. That means that as long as the video project doesn't change and a new tag is not clicked, all frames tagged will go to the current tag regardless of if the tag is the current selected item or not. This will make it easier to tag frames while watching an analysis for example. 

- Create a new tag with Ctrl+t.
- Tag the current frame with 't'. If a current tag is not selected open the create tag dialog.
- Untag the current frame with 'r'.
- The next/prev poi buttons jump to the appropriate tag. 

This one also changes how tags and look and how they work and treeitemclicked is no longer recursive.


ViAn/GUI/Analysis/analysisslider.cpp
```diff
@@ -31,11 +31,19 @@ void AnalysisSlider::paintEvent(QPaintEvent *ev) {
     //Get one frame's width of the slider
     double c = (double)(groove_rect.right()-groove_rect.left())/maximum();
```
fredrikiselius: Should have a more descriptive name than `c`. Perhaps `frame_width` or similar. I know that this wasn't a new addition but we should still change it.

ViAn/GUI/Analysis/analysisslider.cpp
```diff
@@ -31,11 +31,19 @@ void AnalysisSlider::paintEvent(QPaintEvent *ev) {
...
 
-    QBrush brush = Qt::yellow;
+    QBrush brush = Qt::red;
 
-    // Draws the detections on slider and the tags
-    if ((m_show_pois||m_show_tags)&& show_on_slider) {
-        if(m_show_tags) brush = Qt::red;
+    // Draws the tags
+    if (m_show_tags && show_on_slider && m_tag) {
```
fredrikiselius: I think it would be good with a comment that clarifies what we check for here

ViAn/GUI/TreeItems/tagframeitem.cpp
```diff
@@ -0,0 +1,20 @@
...
+}
+
+void TagFrameItem::set_state(TagFrame* t_frame) {
+    m_t_frame = t_frame;
+}
+
+void TagFrameItem::remove(){}
+
+void TagFrameItem::rename() {
+    // TODO
```
fredrikiselius, line 15: Is this fixed in any of the other branches? Just so that we won't forget

ViAn/GUI/projectwidget.cpp
```diff
@@ -132,22 +132,27 @@ void ProjectWidget::start_analysis(VideoProject* vid_proj, AnalysisSettings* set
  * @param tag
  * Adds a tag 'tag' under vid_proj
  */
-void ProjectWidget::add_basic_analysis(VideoProject* vid_proj, BasicAnalysis* tag) {
+void ProjectWidget::add_basic_analysis(VideoProject* vid_proj, Tag* tag) {
```
fredrikiselius: We should probably change the function name so it matches what is added. `add_tag` perhaps

ViAn/GUI/projectwidget.cpp
```diff
@@ -156,6 +161,62 @@ void ProjectWidget::add_basic_analysis(VideoProject* vid_proj, BasicAnalysis* ta
     vid_item->setExpanded(true);
 }
 
+/**
+ * @brief ProjectWidget::add_frames_to_tag
+ * Create the TagFrameItems from all the frames. Used when opening a project
+ * @param item
+ */
+void ProjectWidget::add_frames_to_tag(TreeItem* item) {
```
fredrikiselius: Minor detail but we should change the function name to `add_frames_to_tag_item` so it is clear that the frames are added to a tree item and not a tag object

ViAn/GUI/projectwidget.cpp
```diff
@@ -156,6 +161,62 @@ void ProjectWidget::add_basic_analysis(VideoProject* vid_proj, BasicAnalysis* ta
...
+    }
+}
+
+/**
+ * @brief ProjectWidget::add_new_frame_to_tag
+ * Create a new TagFrameItem for the newly tagged frame
+ * @param frame
+ * @param t_frame
+ */
+void ProjectWidget::add_new_frame_to_tag(int frame, TagFrame* t_frame) {
```
fredrikiselius: Same here as with `add_frames_to_tag`

ViAn/GUI/projectwidget.cpp
```diff
@@ -156,6 +161,62 @@ void ProjectWidget::add_basic_analysis(VideoProject* vid_proj, BasicAnalysis* ta
...
+    }
+    m_tag_item->addChild(tf_item);
+}
+
+/**
+ * @brief ProjectWidget::remove_frame_from_tag
+ * Remove the frame from the current tag
+ * @param frame
+ */
+void ProjectWidget::remove_frame_from_tag(int frame) {
```
fredrikiselius: Same thing here with the function name

ViAn/Project/Analysis/tag.cpp
```diff
@@ -1,112 +1,91 @@
...
 }
 
-void Tag::add_interval(AnalysisInterval *an_interval){
-    BasicAnalysis::add_interval(an_interval);
-    merge_intervals();
+void Tag::remove_frame(int frame) {
+    auto it = tag_map.find(frame);
+    if (it != tag_map.end()) {
+        tag_map.erase(it);
+        delete (*it).second;
```
fredrikiselius, line 23: Is this working ok? [Here](http://en.cppreference.com/w/cpp/container/map/erase) it says that the iterator will be invalidated after the erase

ViAn/GUI/TreeItems/tagframeitem.cpp
henpe071: Currently not fixed, but also not sure if wanted. The tag frames have the frame number as name right now. Do we wanna give the user the option to change that name? If we do, it'll come in a later pull request.

ViAn/Project/Analysis/tag.cpp
henpe071: I can look some more into this.

ViAn/GUI/TreeItems/tagframeitem.cpp
fredrikiselius: Ah, didn't really think about that. Frame number should be enough then.

ViAn/GUI/projectwidget.cpp
henpe071: Once upon a time it had that name as seen in the comment

ViAn/GUI/TreeItems/tagframeitem.cpp
henpe071: Can also say that right now the user can change the name of them but the name should not be saved. We might wanna lock the name field so it can't be changed especially if we don't save the new name.

Issue #41, opened 2018-04-20
Display bugs when loading video with stored state


fredrikiselius:
* First frame is displayed before jumping to stored frame
* Playback slider jumps past the stored number and then back

Issue #34, opened 2018-04-13
Drawings drawn before frame


fredrikiselius:
When jumping between frames containing drawings, the drawings are displayed before the actual frame change. Minor detail but should be fixed

PR #51, opened 2018-05-30
Tag save state

henpe071:
Tags save the state of the video when it was created. The state contains the frame, scale factor and anchor. This is used to jump to the exact location and state of the tag. This branch is a continuation of Tags-and-state.
Videos now also remembers the state they were in when changing videos. 

Issues #34 & #41 are, if not removed, at least less noticable.

Continuation of #50 

TODO
Next and prev poi buttons have not yet been updated to load the state of the tags.


ViAn/GUI/framewidget.cpp
```diff
@@ -106,8 +106,9 @@ void FrameWidget::paste() {
     update();
 }
 
+// TODO Not need?
 void FrameWidget::set_scroll_area_size(QSize size) {
```
fredrikiselius, line 110: Might be better to just remove this then

ViAn/GUI/projectwidget.cpp
```diff
@@ -187,14 +189,16 @@ void ProjectWidget::add_tag(VideoProject* vid_proj, Tag* tag) {
 void ProjectWidget::add_frames_to_tag_item(TreeItem* item) {
```
fredrikiselius, line 189: The same thing is done in each branch of the if statement (except casting). I think it would be better if we create a tag pointer before the statement so that only the casting is necessary inside and then do the rest outside

ViAn/GUI/videowidget.cpp
```diff
@@ -1104,12 +1109,26 @@ void VideoWidget::set_draw_area_size(QSize s) {
...
  * @brief VideoWidget::on_zoom_out
- * Tells the frame processor to decrease zoom
+ * Tells the frame processor to change the zoom with a constant
  */
 void VideoWidget::on_step_zoom(double step){
     update_processing_settings([&](){z_settings.zoom_factor *= step;});
 }
 
+/**
+ * @brief VideoWidget::set_anchor
```
fredrikiselius, line 1119: An update of the comment might be needed here ;)

PR #52, opened 2018-05-30
Prev next poi btns

henpe071:
Added the functionality to jump to next/prev poi with the btns. These buttons will now jump to the correct state when a tag is selected the same way clicking them in the project widget would do.

Continuation of #51, Tag-save-state.


fredrikiselius (commented):
Looks fine

PR #55, opened 2018-05-31
Small bookmarks fixes and some code cleaning

henpe071:
Added the current time in the description.
Changed the create and edit bookmark windows. It no longer uses the custom dialog.
Added Ctrl + enter as a shortcut to close the edit description windows.
The windows now have focus when opened. This make it easier to write the description as you don't have to select the dialog first.
Removed a way for bookmarks to be created even if cancel was clicked.
If the description is empty it won't show the "Description:" in the tool tip and probably won't write it in the report.

Removed the include for custom-dialog and using std namespace which required some fixing.

Most of the important changes are in the bookmark files the other are mostly adding an include or adding "std::".

This branch is a continuation from #52, the prev/next poi branch. All the changes from that specific branch might not be necessary but I wanted the functionality of loading a video with a state for the bookmarks since they will eventually save the state just like the tags. Loading from state isn't added in this branch yet but i didn't wan't a too big pullrequest, especially with all the changes the custom dialog required.


ViAn/GUI/Bookmark/bookmarklist.cpp
```diff
@@ -38,7 +39,7 @@ QListWidgetItem *BookmarkList::get_clicked_item() {
  * Stores the name of the parent container that the widget resides in
  * @param name : name of parent container
  */
-void BookmarkList::set_parent_name(string name) {
+void BookmarkList::set_parent_name(std::string name) {
```
fredrikiselius: More efficient to use `const std::string& name` as function parameter

ViAn/GUI/Bookmark/myinputdialog.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <QDebug>
+
+MyInputDialog::MyInputDialog(QWidget *parnet) : QInputDialog(parnet) {
+    setOption(QInputDialog::UsePlainTextEditForTextInput);
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+}
+
+void MyInputDialog::keyPressEvent(QKeyEvent *e) {
+    if (e->key() == Qt::Key_Return && e->modifiers() & Qt::ControlModifier) {
+        qDebug() << "value" << this->textValue();
```
fredrikiselius: Remove debug

PR #57, opened 2018-06-15
Fix drawing tag

henpe071:
Now jumps to saved frame and update in context menu functions correctly.

### DIFF
--- a/ViAn/GUI/projectwidget.cpp
+++ b/ViAn/GUI/projectwidget.cpp
@@ -804,36 +804,33 @@ void ProjectWidget::context_menu(const QPoint &point) {
  */
 void ProjectWidget::drawing_tag() {
     VideoItem* vid_item;
-    Tag* tag;
-    if (selectedItems().front()->type() == VIDEO_ITEM) {
-        // Create tag drawing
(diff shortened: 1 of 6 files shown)


ViAn/GUI/projectwidget.cpp
```diff
@@ -804,36 +804,33 @@ void ProjectWidget::context_menu(const QPoint &point) {
...
-        tag = item->get_tag();
-        item->takeChildren();
-        tag->clear_intervals();
+        vid_item->get_video_project()->remove_analysis(item->get_tag());
+        delete selectedItems().front();
+    } else if (selectedItems().front()->type() == VIDEO_ITEM) {
+        vid_item = dynamic_cast<VideoItem*>(selectedItems().front());
     }
+    // Create tag drawing
+    Tag* tag = new Tag();
```
fredrikiselius: Not really necessary and not exactly part of the changes here but I think it would look better if the Tag class constructor took a name parameter as well as a parameter for whether its a drawing or not. I leave it up to you

henpe071:
Added a constructor for the tag.

PR #60, opened 2018-06-17
Interpolation switch

fredrikiselius:
Added a checkbox that toggles between bicubic (checked) and nearest neighbor (unchecked) interpolation.
The setting is only applied when the zoom level is above 100%, at lower leves bilinear is used (As before).


henpe071:
Looks good

Issue #48, opened 2018-05-17
Clean up select and none tool


henpe071:
Remove old code

henpe071:
Done

PR #64, opened 2018-06-26
New menus

henpe071:
Moved a lot of things in the menus and toolbars. Added some new icons and overall cleaned up the GUI a little.

- Made a new menu called export. Will later add some more functionality which will be here.
- Remove project are moved to the Recent project dialog.
- Recent project dialog can be accessed from the file menu.
- Added a new button in the main toolbar for opening the project folder.
- Not much else in terms of new functionality. Mostly graphical changes.

All changes are per request for how they want the GUI to look. Just look throught it and make sure it looks ok. A lot of code are just moved.

#48 Is fixed. 

(The perfect +316/-316)


ViAn/GUI/Toolbars/drawingtoolbar.cpp
```diff
@@ -1,93 +1,77 @@
...
     rectangle_tool_act = new QAction(QIcon("../ViAn/Icons/box.png"), tr("Rectangle tool"), this);
     circle_tool_act = new QAction(QIcon("../ViAn/Icons/circle.png"), tr("Circle tool"), this);
+    arrow_tool_act = new QAction(QIcon("../ViAn/Icons/arrow.png"), tr("Arrow tool"), this);
     line_tool_act = new QAction(QIcon("../ViAn/Icons/line.png"), tr("Line tool"), this);
+    pen_tool_act = new QAction(QIcon("../ViAn/Icons/pen.png"), tr("Pen tool"), this);
     text_tool_act = new QAction(QIcon("../ViAn/Icons/text.png"), tr("Text tool"), this);
-    edit_tool_act = new QAction(QIcon("../ViAn/Icons/edit.png"), tr("Edit tool"), this);
-    select_tool_act = new QAction(QIcon("../ViAn/Icons/select.png"), tr("Select tool"), this);
+    color_label = new QLabel();
+    pixmap = new QPixmap(10, 20);
```
fredrikiselius: 10 and 20 should probably be moved to some constants

ViAn/GUI/mainwindow.cpp
```diff
@@ -579,6 +595,24 @@ void MainWindow::gen_report() {
     emit set_status_bar("Generating report. Please wait.");
 }
 
+void MainWindow::init_rp_dialog() {
+    rp_dialog = new RecentProjectDialog(this);
+    rp_dialog->setAttribute(Qt::WA_DeleteOnClose);
+
+    // Create a new project if user presses "new project" or if dialog is rejected
+    connect(rp_dialog, &RecentProjectDialog::new_project, project_wgt, &ProjectWidget::new_project);
+    //connect(rp_dialog, &RecentProjectDialog::rejected, project_wgt, &ProjectWidget::new_project);
```
fredrikiselius: What's the reason for removing this?

ViAn/Project/project.cpp
```diff
@@ -110,7 +110,7 @@ void Project::set_name_and_path(const std::string& name, const std::string& path
     m_name = name;
     if (!path.empty()){
         // Update path for all VideoProjects
-        m_dir = path + "/" + name + "/";
+        m_dir = path + name + "/";
```
fredrikiselius, line 116: Will path always end with a `/`? Otherwise it might be worth checking that it does

fredrikiselius (commented):
It looks good, just some minor things. I haven't been able to try it out yet since I'm on my laptop. I will update with further comments later if I have any after running it

ViAn/GUI/mainwindow.cpp
henpe071: The temp project is now created directly in mainwindow constructor (line 53) so the recent dialog doesn't have to create it on close. This was wanted because now the recent project dialog is accessible in in file menu and I don't want to call new_project() when that dialog is closed.

ViAn/Project/project.cpp
henpe071: I needed to remove a '/' to get the open folder function to work. My guess is that the path saved in the code and memory weren't identical. Currently an extra '/' is added in projectdialog ok_btn_clicked so it will always have at least one. And the functions work when it has multiple.

fredrikiselius:
I've now tested the new changes. It looks really good! I have a couple of thoughts though.

* "Recent project..." in the file menu should probably be changed to "Recent projects.." as to indicate that it won't just open the most recent project.
* The new button for opening the project folder should have it tooltip changed to "Open project folder" so it's more clear which folder it will open.

henpe071:
I'll add an 's'.
The button have gotten some new text in the #71 , glance at that one as well.

fredrikiselius:
Great! Then it can be merged after the conflicts has been solved

PR #66, opened 2018-06-28
Zoom label editable

henpe071:
Zoom label is editable so the user can set the zoom level manually. It accepts values from 1 - 10000 and resets the value on bad input.


fredrikiselius:
Looks fine! Haven't been able to test it yet so will wait with merging until I have

fredrikiselius:
There is a missing input case for "0", "%" and "0%" which crashes the program

henpe071:
I fixed the crash when entering 0. "%" is right now treated as 0 since all '%' are removed from the string.

PR #71, opened 2018-07-02
Shortcuts

henpe071:
Added shortcuts and changed the menus some more according to requests.
- 'b' to bookmark without description
- '*' & '/' to change speed slider
- 'Ctrl H' to reset zoom
- 'N' to toggle interpolate
- 'i' & 'o' to set interval
- 'f' to set focus in the frame label
- 'Ctrl i' import video ("add video" before)
- 'Ctrl alt A' full analysis
- 'Alt 1-7' to set the different drawing tools in order
- 'ctrl r' open recent project
- 'shift g' color correction dialog ("contrast brightness" before)
- 'r' & 'l' rotate
- 'f1' manual (no manual yet tho...)
- 'home' & 'end' jump to first/last frame
- 'x' export frame
- 's' strop button
- 'e' set edit tool
- 'v' set pan/zoom tool
- 'esc' now closes bookmark dialog
- 'alt left/right' jump a page (10 frame) on the slider
- cleaned some tooltips and other things in the menus
- new Edit tool icon

Continuation of #64


ViAn/GUI/Bookmark/bookmarkdialog.h
```diff
@@ -0,0 +1,19 @@
...
+#include <QInputDialog>
+
+class BookmarkDialog : public QInputDialog {
+    Q_OBJECT
+
+public:
+    explicit BookmarkDialog(QWidget* parent = 0);
+    void keyPressEvent(QKeyEvent *e) override;
+
+signals:
```
fredrikiselius: If there are no signals or slots then `signals:` and `public slots:` can be removed

ViAn/GUI/mainwindow.cpp
```diff
@@ -25,11 +25,11 @@
  * @param parent a QWidget variable
  */
 MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent){
-    QDockWidget* project_dock = new QDockWidget(tr("Projects"), this);
-    QDockWidget* bookmark_dock = new QDockWidget(tr("Bookmarks"), this);
-    QDockWidget* drawing_dock = new QDockWidget(tr("Drawings"), this);
-    queue_dock = new QDockWidget(tr("Analysis queue"), this);
-    ana_settings_dock = new QDockWidget(tr("Analysis settings"), this);
+    QDockWidget* project_dock = new QDockWidget(tr("&Projects"), this);
```
fredrikiselius: Why put `&` before the name? It's shown in the title of the widget, is this wanted?

henpe071: The '&' is used to set the ALT shortcuts to navigate the menus. The letter after the '&' will be the shortcut, as seen by that letter being underlined in the menus when pressing ALT. The '&' here was supposed to set the shortcuts in the windows menu and while that works it apparently changed the name of the dockable as well. We don't want that. Can see if I can fix that.

fredrikiselius: I see, thought that was for qactions only.

henpe071: I think thats kinda the problem, the things in the windows menu aren't actions so I have to set the '&' somewhere else which might cause problems.

PR #72, opened 2018-07-03
Thumbnails for videos

henpe071:
Videos in the project tree have thumbnails.


ViAn/GUI/TreeItems/videoitem.cpp
```diff
@@ -24,5 +26,31 @@ VideoProject* VideoItem::get_video_project() {
...
+
+/**
+ * @brief VideoItem::set_thumbnail
+ * Create a thumbnail from the video and set it as icon.
+ */
+void VideoItem::set_thumbnail() {
+    std::string path = m_vid_proj->get_video()->file_path;
+    cv::VideoCapture cap(path);
+    cv::Mat frame;
+    cap >> frame;
```
fredrikiselius: I think this could cause a problem if the program is unable to open the path. The VideoCapture has a isOpened function, we should use it to assert that the file is actually open before attempting to read from it

Something like this (or perhaps using a default thumbnail):
`if (!cap.isOpened()) return;`

Issue #74, opened 2018-07-04
Color corrections on nothing


fredrikiselius:
Color corrections can be used even when a video is not loaded, this causes a crash. (Detected on #71 but was probably introduced much earlier)

PR #75, opened 2018-07-05
Sync & contrast/brightness fixes

### LINKED ISSUE #70: Race condition when loading video with set state
If the thread handling the playback takes a while to load a new video then the state will be set in the processing thread first which will then be defaulted on load completion. This results improper loading when clicking on a image in an image sequence among other things..

henpe071:
Contrast brightness is saved in the video state and therefore remembered when changing between videos. It's also saved in the tags and bookmarks.

This also fixes the sync problem when setting a state and loading a video.
The video info is loaded before the state are set which means that the slider will be in the correct place when clicking on a tagframe item before a video is loaded and prevent the crash in the bookmarks when clicking it before loading a video.

Added a fix for #74 
Fixes #70


ViAn/GUI/videowidget.cpp
```diff
@@ -993,6 +1002,11 @@ void VideoWidget::on_video_info(int video_width, int video_height, int frame_rat
     playback_slider->setMaximum(last_frame);
     set_total_time((last_frame + 1) / frame_rate);
     set_current_time(frame_index.load() / m_frame_rate);
+
+    video_loaded.store(true);
```
fredrikiselius: Is this the same atomic as in videoplayer (m_video_loaded)? If so it shouldn't be set here? Both video_loaded and new_frame should be updated by the videoplayer when it actually has something new otherwise it will probably happen in the wrong thread

henpe071: Yes. This was among the things i tried. Since I don't want these variables to be set before the video is actually loaded and all variable from that is set. But it doesn't seems to be necessary.

PR #77, opened 2018-07-06
Poi toggle fixes

henpe071:
Dechecked when it should and now stops at the last poi frame while playing a video.

### DIFF
--- a/ViAn/GUI/Analysis/analysisslider.cpp
+++ b/ViAn/GUI/Analysis/analysisslider.cpp
@@ -153,6 +153,7 @@ void AnalysisSlider::set_analysis_proxy(AnalysisProxy *analysis) {
     if (analysis) {
         m_ana_interval = std::make_pair(analysis->settings->interval.first, analysis->settings->interval.second);
         rects = analysis->m_slider_interval;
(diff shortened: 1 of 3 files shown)


ViAn/GUI/Analysis/analysisslider.cpp
```diff
@@ -153,6 +153,7 @@ void AnalysisSlider::set_analysis_proxy(AnalysisProxy *analysis) {
     if (analysis) {
         m_ana_interval = std::make_pair(analysis->settings->interval.first, analysis->settings->interval.second);
         rects = analysis->m_slider_interval;
+        last_poi_end = rects.back().second;
```
fredrikiselius, line 156: Can rects be empty here? If so wont' the .second cause a crash?

henpe071: I guess but as it is now it seems to be fine. The only time it should be empty is when the analysis didn't find any POIs but as the analysismethod is now it always creates a poi at the beginning so the analysis will never be completely empty.

fredrikiselius: Alright!

PR #78, opened 2018-07-09
Frame rate

henpe071:
Added the number of frames of the video and the fps as two labels. The fps label is changes when the speed slider is changed and the speed slider is fixed to work as intended. Before it didn't correctly raise the speed on the first tic or update the m_cur_speed_step in video player.


fredrikiselius:
Looks good

Issue #19, opened 2018-03-08
Crashes when saved or closed while an analysis is running


fredrikiselius:
The program crashes when it is saved or closed while an analysis is running.

henpe071:
Popups have been added to prevent these crashes.

PR #80, opened 2018-07-12
Prevent crash when closing project while analysis is running

### LINKED ISSUE #73: Remove queued analysis
There is a remove option for queued analysis items in the project widget that crashes the program

henpe071:
Removed some options from the context menu before the analysis is done.
Now when the project would be closed it prompts the user if it wants to abort all the current running analyses, if not just cancel.

fix #73 and #19 .

### DIFF
--- a/ViAn/GUI/Analysis/analysiswidget.cpp
+++ b/ViAn/GUI/Analysis/analysiswidget.cpp
@@ -102,11 +102,11 @@ void AnalysisWidget::abort_all_analysis() {
 void AnalysisWidget::on_analysis_aborted() {
     analysis_queue.pop_front();
     delete current_analysis_item; // Delete item from tree
-    auto it = abort_map.find(current_method);   
+    auto it = abort_map.find(current_method);
     abort_map.erase(it);
 
     m_queue_wgt->next();
-    if (!analysis_queue.empty()) {        
+    if (!analysis_queue.empty()) {
         current_analysis_item = std::get<1>(analysis_queue.front());
         move_queue();
         perform_analysis(analysis_queue.front());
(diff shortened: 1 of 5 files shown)


niklasniklas:
Works! :-)

ViAn/GUI/Analysis/analysiswidget.cpp
```diff
@@ -94,17 +94,25 @@ void AnalysisWidget::abort_analysis() {
...
+void AnalysisWidget::abort_all_analysis() {
+    abort_all = true;
+    abort_analysis();
+}
+
 void AnalysisWidget::on_analysis_aborted() {
     analysis_queue.pop_front();
     delete current_analysis_item; // Delete item from tree
-    auto it = abort_map.find(current_method);   
+    auto it = abort_map.find(current_method);
```
fredrikiselius: Should only erase the `current_method` is actually found otherwise this could lead to problems 
`if (it != abort_map.end())`

fredrikiselius (commented):
New additions works as they should. I noticed a quite large increase in memory usage after canceling an analysis though. Is this to be expected or do we have a memory leak here? Either way it caused a crash for me, we need to catch the error in the analysis loop

henpe071:
There seems to be some memory that's not cleared when an analysis is removed. What was the crash you got? Works for me.

fredrikiselius:
I have quite limited memory on the computer I'm currently on so when I started two analyses and canceled the first on there wasn't enough memory left for opencv to allocate for the next video so it threw a memory allocation error. Either way it needs to be solved. Should we make it part of this branch or add it as another issue and fix it in another?

henpe071:
I tried with some other videos and with one of them I get memory error even when i queue it with another video. There can definitely be some things done on the analysis code but again I think we should do it in a seperate branch when it's not directly releted to the pull request.

fredrikiselius:
I'll create another issue then (#91) and merge this one

PR #82, opened 2018-07-12
Multiple widgets NEW

henpe071:
It's possible to open multiple seperate videowidgets. To do so, rightclick the video you wanna open and choose "open video in widget" and a new widget will appear. This is not limited to only one new video widget.

I reopen a new one. Easier this way


niklasniklas:
Issue: Close all multiple-widgets when opening other project.

ViAn/GUI/videowidget.cpp
```diff
@@ -158,7 +177,13 @@ void VideoWidget::init_frame_processor() {
     f_processor = new FrameProcessor(&new_frame, &settings_changed, &z_settings, &video_width,
                                      &video_height, &new_frame_video, &m_settings, &v_sync, &frame_index, &o_settings, &overlay_changed);
 
-    QThread* processing_thread = new QThread();
+    try {
+        processing_thread = new QThread(this);
```
fredrikiselius: Shouldn't the following block of code (line 187-212) also be inside the try block? There is no point in connecting if the thread cannot be created

fredrikiselius:
- If a video is playing in a floating widget and the project is closed the program crashes.
- The interval buttons are still enabled in new widgets. Is this wanted?
- The _export frame_ button does nothing in a floating widget

niklasniklas:
Works fine! :-)

fredrikiselius:
Can't play in new widgets anymore. The program instantly freezes when I attempt to. The issues is in the `load_marked_video_state` function. Changing the first condition to `!vid_proj->is_current() || m_vid_proj == nullptr` solves it.

The issue
> If a video is playing in a floating widget and the project is closed the program crashes.

still happens _sometimes_.

I also noticed that some wierd behaviour can occur if a video is opened in the fixed/central widget and in a floating one. If the video in the central widget is at some frame, say 10, and the floating is then played/moved to another frame the shared state will be updated. If I then tag the current frame in the central widget (still 10), the program will create a tag for the frame that the second player is at. I don't expect that we will have time to fix this, just thought I should mention it.

PR #84, opened 2018-07-17
Fixed play/pause button

henpe071:
Play/pause button now have space as shortcut and it only triggers once when clicked.


niklasniklas:
Works! :-)

PR #85, opened 2018-07-18
Delete focus

henpe071:
Pressing the shortcut delete now deletes different things depending on what widgets has focus. In the project widget it deletes the selected item, in drawing widget or video widget it deletes the current drawing with the same conditions as before and in bookmark widget (list) it deletes the selected bookmark or category. 

In bookmark widget the shortcut f2 now opens the change change destription dialog.
In project widget the shortcut f5 is added for updating the current drawing tag.
New dialog when changing the title of a category.

### DIFF
--- a/ViAn/GUI/drawingwidget.cpp
+++ b/ViAn/GUI/drawingwidget.cpp
@@ -378,7 +378,7 @@ void DrawingWidget::delete_item() {
         if (f_item->get_frame() == m_overlay->get_current_frame()) {
             QMessageBox msg_box;
             msg_box.setIcon(QMessageBox::Warning);
-            msg_box.setText("Deleting frameitem\n"
+            msg_box.setText("Deleting drawings on frame "+QString::number(f_item->get_frame())+"\n"
                             "This will delete all drawings on this frame");
             msg_box.setInformativeText("Do you wish to continue?");
             msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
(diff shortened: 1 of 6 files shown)


niklasniklas:
Works! :-)

ViAn/GUI/drawingwidget.cpp
```diff
@@ -375,6 +376,16 @@ void DrawingWidget::delete_item() {
     if (item->type() == FRAME_ITEM) {
         FrameItem* f_item = dynamic_cast<FrameItem*>(item);
         if (f_item->get_frame() == m_overlay->get_current_frame()) {
+            QMessageBox msg_box;
+            msg_box.setIcon(QMessageBox::Warning);
+            msg_box.setText("Deleting frameitem\n"
```
fredrikiselius: `Deleting frameitem\n` should probably be renamed into something like `deleting drawings from frame` or similar since the user won't know what a frameitem is

fredrikiselius (commented):
The new additions works fine but there are some issues concerning deletion order. 
When deleting a video its bookmarks will remain and attempting to delete them will cause a crash (Same when double clicking them).

henpe071:
That functionality has never really been added. Currently when a video project is deleted is does nothing with the bookmarks. While this should definitely be fixed, it also isn't really what this pull request is about. As the time is running out we should instead prioritize getting in the code that works so the pull requests doesn't get too big and make a new one/issue for this.

fredrikiselius:
Sounds reasonable! Created an issue, merging this one now

PR #88, opened 2018-07-23
Sync fixes & zoom overhaul

fredrikiselius:
This became quite large and covers various things, sorry about that..

**Sync:**

_I've moved around some locks and other thread managing stuff. It seems to work fine on my computer but try to switch around between videos/tags/frames just to be sure there are no dead locks_
- Refactored video load so that the frame index that is stored in the state is loaded directly instead of after frame zero
- Removed the extra processing call that was made for the overlay when loading a new video. All state information is now loaded in the _load_video_ part of the frame processor

- Added additional locks to _load_marked_video_state_

**Zoom/Frame processing:**
_Large overhaul of the Zoomer. It no loger uses the zoom rectangle but instead a rectangle identical in size to the viewport. These changed were made since image sequences barely worked with the old implementation_

- Rotation is now done properly and should rotate around the central point in the viewport. (Does not rotate drawings properly yet)
_Rotation causes lag on very large frames.._
- Resizing the window will not move the video center (don't know if this worked before or not)
- When zooming using the drawn rectangle the Zoomer will focus on it's center and use the longest side to set the scale factor (rectangle is still locked to viewport size but it should be easy to fix)
- Rotation is stored in the state

& Various minor changes


niklasniklas:
Sync:
Works ok. Tested tags, different videos, different zoomstate and color corrections. Recived no deadlocks. 

Zoom:
Rotation of video and resizing window works ok. 
Rotation also ok i stored state (with tags and bookmarks)

> When zooming using the drawn rectangle the Zoomer will focus on it's center and use the longest side to set the scale factor (rectangle is still locked to viewport size but it should be easy to fix)

Here are som issues with the Zoomer:
The red line in the zooming rectangle shouldn't be visiable. When zooming with rectangle the part inside the marked area i.e. the green-rectangle should always be visible when zooming is complete. The zoom should centre round the green-rectangle area, calculate if the horizontal och vertical is the one that is limiting the zoomfactor. (More details sent i email)

There are also som issues when zooming with "+" key or Ctrl+mouse wheel. The canvas that is shown seem to be centrered and crops some parts of the video/image-sequence. (Tested scenario in mail)

henpe071:
Similar comments as Niclas. Sync seems to work (on videos). Zoom and rotation mostly works on what I've tested. But if you zoom out in another place than you zoom in the canvas is cropped. Easy to see when ctrl zoom in in one corner and ctrl zoom out in another corner.

The zoomrectangle's width is always the one checked which gives the wrong zoom when drawing a tall pillar with the tool.

Then I have some other things I found related to image sequence. 
- First of all, if a sequence is named with "åäö" it won't read the images. 
- The first image clicked in a sequence will determine the zoom factor for all images in the sequence. This looks weird if the first image is really small and therefore sets a large zoomfactor. Though, this might be tricky to fix while still having the state work.
- I had some trouble with loading imagesequences after saving. New project->import sequence->save proj->cant view images.
- Right now if you close a project with an img seq to create a new project, when clicking a video in the new project the last clicked image from the last project is quickly shown. I even got i to stick one time. So the image from the old project was loading insead of the clicked image from the new project.

fredrikiselius:
@niklasniklas 

- [x] The red line in the zooming rectangle shouldn't be visiable.
- [x] When zooming with rectangle the part inside the marked area i.e. the green-rectangle should always be visible when zooming is complete. The zoom should centre round the green-rectangle area, calculate if the horizontal och vertical is the one that is limiting the zoomfactor
- [x] There are also som issues when zooming with "+" key or Ctrl+mouse wheel. The canvas that is shown seem to be centrered and crops some parts of the video/image-sequence.

@henpe071 

> Sync seems to work (on videos).

Not on sequences?

- [x]  if you zoom out in another place than you zoom in the canvas is cropped. Easy to see when ctrl zoom in in one corner and ctrl zoom out in another corner. 
- [x] The zoomrectangle's width is always the one checked which gives the wrong zoom when drawing a tall pillar with the tool.

The issues with image sequences will be solved in another branch.

niklasniklas:
**Zoom in video**
Zoom on video OK! :-) 
Also ok with zoom with + and ctrl+mouse wheel

**Some issues when zooming img sequence:**
Sometimes the images in a sequence takes the size of some other image.  The image sometimes seem to have the accual size of the first image and it cannot be panned. 

_Bug can be reproduced:_ 
* New project
* Loaded three sequences. 1) tiff images (from pumparna) 2) jpgs 3) mixture of 1 and 2 (where the first img in seq is a small jpg)
* When selecting an img from seq 1) and then selecting the same tiff image in 3) the image receives size of the first image in 3) and cannot be panned. 

The issue only seems to appear once with jpgs. But after you have panned a jpg image then the bug never appears again. 
Clicking around between images in seq 1), 2) and 3) in different order seem to receive different sizes on images. Mostly the error can be seen on the tiff images in seq 3) seem when you go from seq 1) to seq 3). Sometimes the images can be panned and sometimes they cannot. 


**Image sequence not showing**
I also got Henriks bug: "New project->import sequence->save proj->cant view images.". 
Bug does not appear if the project is saved at least once before "save proj". Could it be som issue with copying/moving the "image sequence" files from temporary folder?

ViAn/ViAn.pro
```diff
@@ -1,276 +1,274 @@
...
-
-#
-# ANALYSIS
-#
-SOURCES += Analysis/analysismethod.cpp\
-    Analysis/motiondetection.cpp
-
-HEADERS += Analysis/motiondetection.h
+#-------------------------------------------------
+#
```
henpe071, line 2: Why is this whole file marked as new when only zoomrect is removed?

ViAn/ViAn.pro
```diff
@@ -1,276 +1,274 @@
...
+    LIBS += -lopencv_videoio
+    LIBS += -lopencv_highgui
+}
+
+#
+# START: FILEHANDLER
+#
+SOURCES += \
+    Filehandler/saveable.cpp \
+
```
henpe071, line 231: What is this "^M"?

ViAn/Video/videoplayer.cpp
```diff
@@ -204,16 +204,31 @@ bool VideoPlayer::wait_load_read(){
...
     // Read new frame and notify processing thread
     {
-        std::lock_guard<std::mutex> lk(m_v_sync->lock);
+        m_v_sync->lock.lock();
+        int load_video_at = m_v_sync->frame_index_on_load;
+        m_frame->store(load_video_at);
         load_video();
-        if (!m_capture.read(m_v_sync->frame)) {
+        if (load_video_at != 0) {
+            m_frame->store(load_video_at);
```
henpe071: Do you really need to store in m_frame twice here?

ViAn/Video/videoplayer.cpp
```diff
@@ -204,16 +204,31 @@ bool VideoPlayer::wait_load_read(){
...
             m_is_playing->store(false);
             playback_stopped();
-
+            m_v_sync->lock.unlock();
             return false;
         }
-        m_new_frame->store(true);
     }
-    m_v_sync->con_var.notify_one();
+    m_v_sync->con_var.notify_all();
```
henpe071, line 234: right now, only the else if reach this, is this wanted?

ViAn/Video/zoomer.cpp
```diff
@@ -3,7 +3,7 @@
 #include <QDebug>
 
 Zoomer::Zoomer() {
-    anchor = QPoint(0,0);
+//    anchor = QPoint(0,0);
```
henpe071: Remove

ViAn/Video/frameprocessor.cpp
```diff
@@ -152,6 +153,24 @@ void FrameProcessor::process_frame() {
...
+/**
+ * @brief FrameProcessor::load_zoomer_state
+ */
+void FrameProcessor::load_zoomer_state() {
+    m_z_settings->set_state = false;
+    m_zoomer.load_state(m_z_settings->center, m_z_settings->zoom_factor, m_z_settings->rotation);
+
+    if (m_z_settings->rotation == 0) {
+        m_rotate_direction = ROTATE_NONE;
+    } else if (m_z_settings->rotation == 90) {
```
henpe071: Make these constants. Maybe make use of the ROTATE_90 etc constants we already have and make a constant for 90°.

ViAn/Video/frameprocessor.cpp
```diff
@@ -297,6 +306,21 @@ void FrameProcessor::update_overlay_settings() {
     }
 }
 
+void FrameProcessor::update_rotation(const int& direction) {
+    if (direction != m_rotate_direction) {
+        m_rotate_direction = direction;
+        if (m_rotate_direction == ROTATE_NONE) {
+            m_zoomer.update_rotation(0);
+        } else if (m_rotate_direction == ROTATE_90) {
```
henpe071: same here, constants

ViAn/Video/zoomer.cpp
```diff
@@ -43,204 +80,291 @@ void Zoomer::set_zoom_rect(QPoint p1, QPoint p2) {
...
+ * the angle difference between the given and current angle.
+ *
+ * Also updates the center position of the viewport rectangle,
+ * since it will change when the uncut frame is rotated.
+ *
+ * Finally it rotates the frame rectangle if necessary
+ * @param angle :   Desired rotation
+ */
+void Zoomer::update_rotation(const int &angle) {
+    double angle_diff{(angle - m_angle) * M_PI /180};
```
henpe071: same here, constant

ViAn/Video/zoomer.cpp
```diff
@@ -43,204 +80,291 @@ void Zoomer::set_zoom_rect(QPoint p1, QPoint p2) {
...
+/**
+ * @brief Zoomer::get_frame_rect
+ * @return cv::Rect :   frame rectangle
+ */
 cv::Rect Zoomer::get_frame_rect() const {
     return m_frame_rect;
 }
 
+/**
+ * @brief Zoomer::get_cutting_rect
```
henpe071: wrong name

ViAn/Video/zoomer.cpp
```diff
@@ -43,204 +80,291 @@ void Zoomer::set_zoom_rect(QPoint p1, QPoint p2) {
...
+    bool left_contained{p[0].x >= 0 || p[1].x >= 0};
+    bool up_contained{p[1].y >= 0 || p[2].y >= 0};
+    bool right_contained{p[2].x <= m_transformed_frame_rect.width || p[3].x <= m_transformed_frame_rect.width};
+    bool down_contained{p[3].y <= m_transformed_frame_rect.height || p[0].y <= m_transformed_frame_rect.height};
+
+    int dx{}, dy{};
+    if ((is_moving_left && left_contained) || (is_moving_right && right_contained)) {
+        dx = x;
+    }
+    if (is_moving_up && up_contained || is_moving_down && down_contained) {
```
henpe071: I have a warning suggesting parentheses here, try to fix warnings as well

fredrikiselius:
@niklasniklas 

- [x] Sometimes the images in a sequence takes the size of some other image. The image sometimes seem to have the accual size of the first image and it cannot be panned.
- [ ] Image sequence not showing (#92)
_If I remember correctly the images are copied from the temporary folder but the stored paths are not updated so the program will continue to look for the images in the old location. I will add a fix in a new branch_

@henpe071 

- [x] The first image clicked in a sequence will determine the zoom factor for all images in the sequence. This looks weird if the first image is really small and therefore sets a large zoomfactor. Though, this might be tricky to fix while still having the state work.
_Made it so that the program will fit to viewport if it encounters a new image size and there is no state to load. For now this only works if the playback controls are used (clicking on a image name in the Images container will set a state..)._

henpe071 (commented):
The code looks fine without getting too deep into the zoomer. I had some comments.

And I know this isn't a new bug but it's possible to crash the program by zooming in alot or zooming out alot (zoom factor around 460k). Thought since you were working with the zoomer you could add a check for this.

ViAn/Video/zoomer.h
```diff
@@ -63,6 +64,8 @@ class Zoomer{
     cv::Rect get_frame_rect() const;
     cv::Rect get_view_rect() const;
 
+    void enforce_frame_boundaries();
```
henpe071: Small thing but i would put this above the setters and getters since it doesnt belong with them

ViAn/ViAn.pro
fredrikiselius: I have no idea.. All the changes done in the .pro file are done automatically

ViAn/ViAn.pro
fredrikiselius: It's some wierd line ending that is not converted for some reason

ViAn/Video/videoplayer.cpp
fredrikiselius: No, will remove the second one. Missed it during cleanup

ViAn/Video/videoplayer.cpp
fredrikiselius: The first check calls on `set_frame` which in turn calls `synced_read` which has a notify so that one should be fine. The program will only get to the last else if it can't read the frame. If there is no new frame we don't want to wake the processing thread so it should also be ok. The last part is how it worked before

niklasniklas:
Zoom issue when switching img-seq workes fine now.

PR #93, opened 2018-07-25
Image sequence path fix

### LINKED ISSUE #92: Image sequence added before save can't be played after
The images are copied to the new save location but the path stored in the program is not updated. Should be solved in the same way that bookmark handles its location updates for thumbnails

#88 

> I had some trouble with loading imagesequences after saving. New project->import sequence->save proj->cant view images.

>  "New project->import sequence->save proj->cant view images.".
Bug does not appear if the project is saved at least once before "save proj". Could it be som issue with copying/moving the "image sequence" files from temporary folder?

> If I remember correctly the images are copied from the temporary folder but the stored paths are not updated so the program will continue to look for the images in the old location. I will add a fix in a new branch

fredrikiselius:
The path to the copied images in sequences is now update correctly when saved. Fixes #92.

I used a tool _dos2unix_ to remove wierd line endings (^M) which made git think everything has changed in the imagesequence file.. The only changes I actually made in that file are in the `the reset_root_dir() `function

### DIFF
--- a/ViAn/GUI/projectwidget.cpp
+++ b/ViAn/GUI/projectwidget.cpp
@@ -1194,6 +1194,15 @@ bool ProjectWidget::save_project() {
             emit proj_path(m_proj->get_dir());
             QDir dir;
             dir.mkpath(QString::fromStdString(m_proj->get_dir()));
+
+            // If the current video is a sequence then it needs to be reloaded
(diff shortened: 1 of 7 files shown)


henpe071 (approved):
Looks good

PR #94, opened 2018-07-27
Thread shutdown fix

fredrikiselius:
This should fix the issue I encountered when opening creating a new project whilst multiple videos were playing. Also added the fix I mentioned in #82.


henpe071:
Perfect

PR #97, opened 2018-08-29
Analysis details fix

### LINKED ISSUE #95: Analysis details always showing
When "analysis details" is enabled and there is no deatils to be shown - the details from the previous analysis is shown.

henpe071:
The analysis details will now be hidden when they should and only be visible when the analysis is clicked from the project-tree and the details settings is checked.

Fixes #95


niklasniklas (approved):
Works fine

PR #101, opened 2018-08-30
Playing a video in a floating widget will no longer update the state

### LINKED ISSUE #96: Shared state in multiple widgets
_I also noticed that some wierd behaviour can occur if a video is opened in the fixed/central widget and in a floating one. If the video in the central widget is at some frame, say 10, and the floating is then played/moved to another frame the shared state will be updated. If I then tag the current frame in the central widget (still 10), the program will create a tag for the frame that the second player is at. I don't expect that we will have time to fix this, just thought I should mention it._

henpe071:
To prevent the issue noted in #82 

Also removed load_marked_video function.

Fixes #96

### DIFF
--- a/ViAn/GUI/drawingwidget.cpp
+++ b/ViAn/GUI/drawingwidget.cpp
@@ -276,7 +276,12 @@ void DrawingWidget::tree_item_clicked(QTreeWidgetItem *item, const int &col) {
     switch (item->type()) {
     case FRAME_ITEM: {
         FrameItem* frame_item = dynamic_cast<FrameItem*>(item);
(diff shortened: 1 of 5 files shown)


niklasniklas:
Issue with zoom. Zoom factor is updated in centralwidget if changed in folating widget.
Testsequence:
* Zoom in floating widget showing film "1"
* Click on film in project tree (other than "1")
* Click on "1" in project tree

niklasniklas (changes requested):
Zoom in floating widget is updated in central widget. See comment...

niklasniklas (approved):

PR #111, opened 2018-09-04
Clear videowidget when project closes

henpe071:
Couldn't reset it like I wanted to so hade to do the backup plan of just clearing some flags. But the analysisinterval is at least cleared from the slider now when the project closes.


niklasniklas (approved):
Works fine

PR #112, opened 2018-09-04
Drawing resolution

henpe071:
Drawings are now scaled and drawn after the Mat is scaled. This gives them a lot better resolution as they are always drawn in the current resolution.

The text drawing are an exception as is functions differently. As it doesn't have a start and end point as the other drawings and just a start and a text size it's not possibly to scale it the same way. So text drawings are still drawn as they used to.

### DIFF
--- a/ViAn/Video/frameprocessor.cpp
+++ b/ViAn/Video/frameprocessor.cpp
@@ -178,20 +178,22 @@ void FrameProcessor::process_frame() {
     //    cv::rectangle(tmp, view_rectangle.tl() * factor, view_rectangle.br() * factor, cv::Scalar(255,255,255), 2);
     //    cv::imshow("test", tmp); // can cause a deadlock
 
+    int frame_num = m_frame_index->load();
+
     // Draws the text drawings on the overlay
-    m_overlay->draw_overlay(manipulated_frame, m_frame_index->load());
+    m_overlay->draw_overlay(manipulated_frame, frame_num);
 
     // Apply zoom
     m_zoomer.scale_frame(manipulated_frame);
 
     // Draws the other drawings on the overlay
-    m_overlay->draw_overlay_scaled(manipulated_frame, m_frame_index->load(), Utility::from_qpoint(m_z_settings->anchor), m_z_settings->zoom_factor);
+    m_overlay->draw_overlay_scaled(manipulated_frame, frame_num, Utility::from_qpoint(m_z_settings->anchor), m_z_settings->zoom_factor);
 
     // Applies brightness and contrast
     m_manipulator.apply(manipulated_frame);
 
     // Emit manipulated frame and current frame number
-    emit done_processing(m_frame, manipulated_frame, m_frame_index->load());
+    emit done_processing(m_frame, manipulated_frame, frame_num);
 }
 
 /**
--- a/ViAn/Video/frameprocessor.cpp
+++ b/ViAn/Video/frameprocessor.cpp
@@ -170,7 +170,7 @@ void FrameProcessor::process_frame() {
         cv::rotate(manipulated_frame, manipulated_frame, m_rotate_direction);
     }
 
-    // Create zoom perview mat
+    // Create zoom preview mat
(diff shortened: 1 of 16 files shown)


niklasniklas:
Looks fine. Works fine with document.
However encounteredd an issue with extreme zooming in. I got an horisontal line. Se images below.

skarmklipp1
skarmklipp2

niklasniklas:
@fredrikiselius could you also look through the codechanges?

niklasniklas (approved):

henpe071:
Those horizontal lines have appeared before, epecially when drawing a circle like a vertical line. I Think they are cause by the coordinates to somehow overlap and cause some trouble. I haven't seen it while zooming like this before but it might be the same problem since when zoomed in that far the coordinated might overlap. I have at least added an issue about it and might look into it more later.

ViAn/Video/frameprocessor.cpp
```diff
@@ -170,20 +170,23 @@ void FrameProcessor::process_frame() {
...
 
-    // Draws the overlay
+    // Draws the text drawings on the overlay
     m_overlay->draw_overlay(manipulated_frame, m_frame_index->load());
 
     // Apply zoom
     m_zoomer.scale_frame(manipulated_frame);
 
+    // Draws the other drawings on the overlay
+    m_overlay->draw_overlay_scaled(manipulated_frame, m_frame_index->load(), Utility::from_qpoint(m_z_settings->anchor), m_z_settings->zoom_factor);
```
fredrikiselius: There is a slight risk that `m_frame_index->load()` could cause an issue here. It's an atomic action and we use it quite freely without locks. Therefore the user could potentially change the frame number after the frame has been loaded and thus the program would draw the wrong drawings. Would probably be good to take a look at.

ViAn/Video/overlay.cpp
```diff
@@ -20,21 +20,46 @@ Overlay::~Overlay() {
...
+ * Draws the text drawings from an overlay on top of the specified frame.
  * @param img Frame to draw on
  * @param frame_nr Number of the frame currently shown in the video.
  */
 void Overlay::draw_overlay(cv::Mat &frame, int frame_nr) {
     if (show_overlay) {
         for (auto it = overlays[frame_nr].begin(); it != overlays[frame_nr].end(); it++) {
-            if ((*it)->get_show()) {
+            // Only draw the text drawings
+            if ((*it)->get_show() && (*it)->get_shape() == TEXT) {
```
fredrikiselius, line 31: Might be more efficient to store the value of `*it` instead of dereferencing twice. Not completely sure though

ViAn/Video/overlay.cpp
```diff
@@ -20,21 +20,46 @@ Overlay::~Overlay() {
...
+ * @param anchor - The top left coordinate of the zoomrect, used to scale the drawings
+ * @param scale_factor - The zoom percent, used to scale the drawings
+ */
+void Overlay::draw_overlay_scaled(cv::Mat &frame, int frame_nr, cv::Point anchor, double scale_factor) {
+    if (show_overlay) {
+        for (auto it = overlays[frame_nr].begin(); it != overlays[frame_nr].end(); it++) {
+            // Don't draw text drawings since they work a little different.
+            // They can't be scaled as big and small as needed the same way as other drawings
+            // They are drawing first and then scaled with the Mat
+            if ((*it)->get_show() && (*it)->get_shape() != TEXT) {
```
fredrikiselius, line 54: Same here with dereferencing

fredrikiselius:
While the changes looks mostly ok as is, I think it's important to look into the thing I mentioned with `m_frame_index->load()`

ViAn/Video/overlay.cpp
henpe071: I don't know what's more fficient but this is such a small thing with minor impact on efficiency that readability should be what's most important. The compiler will handle the code and from what I understand will treat the two options the same. I can change it to store the value if we feel like that is more readable.

PR #121, opened 2018-09-07
Color corrections are now in a dockable widget

henpe071:
The widget looks similar to before. Two sliders, one for contrast and one for brightness. These sliders update the current frame butt will not save the changes. To save the changes to the video state click the "apply"-button.

The default button will reset and apply the color corrections to the default values.


niklasniklas:
Unexpected behaviour:
Ex. 
* Go to a frame (ex. 100)
* Change color and contrast (dont press apply)
* Tag frame ("T")
* The tag doesnt receive the color corrections

niklasniklas:
Unexpected behaviour:
Pressing the pin-button (updating the current tag)
Does not take the current settings on color correction

niklasniklas:
@henpe071 and @klwing021 
Discussion:
Should "Default" also include "Apply"?

henpe071:
> Unexpected behaviour:
>Ex.

>Go to a frame (ex. 100)
>Change color and contrast (dont press apply)
>Tag frame ("T")
>The tag doesnt receive the color corrections

Expected behavior (by me). Since the changes are not applied they will not be used when tagging a frame. When creating a tag it will read the Videostate which have not been updated and still have the old values. This might not be the wanted behavior however, or at least there maybe should be some better indication that there are changes to the frame that are not applied.
This might have to be discussed more how it's supposed to work.

PR #122, opened 2018-09-07
Video overview / zoom preview

### LINKED ISSUE #104: Zoom overview i widget

fredrikiselius:
Added a new widget that contains a minature image of the original video. It will always show the whole video and overlay the zoom rectangle, i.e. the bounding box of the image displayed in the framewidget.

Closes #104 .


niklasniklas (approved):

niklasniklas:
@henpe071 COuld you look through the codechanges?

niklasniklas:
Works fine!

ViAn/GUI/mainwindow.cpp
```diff
@@ -364,6 +375,7 @@ void MainWindow::init_view_menu() {
     view_menu->addAction(toggle_bookmark_wgt);
     view_menu->addAction(toggle_queue_wgt);
     view_menu->addAction(toggle_ana_settings_wgt);
+    view_menu->addAction(toggle_zoom_preview_wgt);
```
henpe071: Maybe a small thing but this line should be moved up two steps so the zoom preview is after the bookmark widget. By looking at the other changes that's where it should be. It also makes the context menu and view (window) menu the identical.

ViAn/GUI/mainwindow.cpp
```diff
@@ -373,6 +385,7 @@ void MainWindow::init_view_menu() {
     toggle_bookmark_wgt->setStatusTip(tr("Show/hide bookmark widget"));
     toggle_queue_wgt->setStatusTip(tr("Show/hide analysis queue widget"));
     toggle_ana_settings_wgt->setStatusTip(tr("Show/hide analysis info widget"));
+    toggle_zoom_preview_wgt->setStatusTip(tr("Show/hide zoom preview"));
```
henpe071: Same here so everything is in the same order.

ViAn/GUI/zoompreviewwidget.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <QDebug>
+
+/**
+ * @brief zoompreviewwidget::center_image
+ * Centers the image in the widget
+ * @param s QSize   : The size of the image to be centered
+ */
+void ZoomPreviewWidget::center_image(const QSize &s) {
+    // Will center the image both horizontally and vertically
+    // anchor = QPoint((this->width() - s.width()) / 2, (this->height()- s.height()) / 2);
```
henpe071: Outcommented line of code?

henpe071:
Aside from those small comments is looks and works fine.

ViAn/GUI/zoompreviewwidget.cpp
fredrikiselius: I saved that because I wasn't sure if we wanted to center the widget vertically as well. It looks a bit weird if the preview widget is the only one in a dock area. @niklasniklas or @klwing021 might be able to answer that?

niklasniklas: @fredrikiselius: I would prefer centered both horizontal and vertical.

niklasniklas: Other than that. Ok to merge after solving conflicts and changing the order in context menue.

henpe071:
Works fine

PR #125, opened 2018-09-07
Apply updates color correction on tag

### LINKED ISSUE #113: Color corrections update current tag when clicking apply
Only update the tag if it is selected in the project tree

henpe071:
If there is a tag on the current frame in the active tag, update that tag's color correction when clicking apply.

Continuation of #121 

Closes #113


niklasniklas:
See comments in #121. 
Other than that Works fine!

PR #132, opened 2018-09-11
Fix the problem with circle drawing drawing horizontal lines

### LINKED ISSUE #126: Horizontal lines when drawing a circle
When drawing a circle like a vertical line, horizontal lines will appear from the top and bottom of the circle and stretch across the framewidget.

henpe071:
Changed so the x coordinates won't overlap to prevent the horizontal lines.

Fixes #126


niklasniklas (approved):
Works fine

niklasniklas:
Unable to test if this also solves the issue encountered with extreme zooming in drawing in screen resolution (see #112 ) since that pull request is still pending

PR #144, opened 2018-09-17
Drawing on wrong frame

### LINKED ISSUE #42: Drawing appear on the wrong frame while playing a video

henpe071:
Videoplayer was processing the frame before updating the frame number, which means the drawings on the overlay were drawn on the wrong frame while video is playing.

Fixes #42


klwing021:
If an object is selected, the bounding-box is still drawn on the wrong frame

niklasniklas:
Works fine!

ViAn/Video/videoplayer.cpp
```diff
@@ -125,10 +125,16 @@ void VideoPlayer::check_events() {
             // Timer condition triggered. Read new frame
             if (m_is_playing->load()) {
                 std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
-                if (!synced_read()) continue;
                 ++*m_frame;
```
fredrikiselius, line 130: We should add a comment here of why we are doing it this way.

PR #145, opened 2018-09-17
Delete analysis fixes

### LINKED ISSUE #129: Removing analyses causes problems and eventual crash
Removing an anaysis will remove the analysis from the folder even if the project isn't saved. This will cause problem the next time the project is opened as the analysis item is still in the project tree but the related analysis file is gone. Now pressing show details causes a crash and the analysis won't show any detection boxes.

### LINKED ISSUE #134: Deleting analysis from project tree
Deleting an analysis during the analysis or a queued analysis crasches the program

henpe071:
No longer able to delete not-yet-done analyses with the delete shortcut. Added a check to prevent it.

When removing an analysis the path to the analysis is saved in a list and the file in the folder in untouched. Later when the project is saved the list of "removed" project is removed from the folder. If the project is not saved, the list is just cleared so the file are still in the folder.

Fixes #129 
Fixes #134

### DIFF
--- a/ViAn/GUI/projectwidget.cpp
+++ b/ViAn/GUI/projectwidget.cpp
@@ -988,6 +988,10 @@ void ProjectWidget::update_settings() {
  * If a folder is selected then it will delete all subitems as well.
  */
 void ProjectWidget::remove_item() {
+    if (!dynamic_cast<AnalysisItem*>(currentItem())->is_finished()) {
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


klwing021:
Remove the file from the folder and "save" that part of the project. Same when a new analysis is created, the analysis will be automatically saved even if the project is closed without saving.

niklasniklas:
Works fine!

PR #149, opened 2018-09-18
Not saved analyses will be removed when the project is closed

### LINKED ISSUE #147: Not removing new analyses when closing project without saving
When a new analysis is created it will add its file to the folder but it will not be removed when the project is closed while not saved.

henpe071:
The analysis file will be removed from the folder when the project is closed when the analysis is not saved.

Fixes #147


niklasniklas:
Works fine!

PR #152, opened 2018-09-20
Speed slider slow down

### LINKED ISSUE #151: Speedslider slows down the entire program
Should only slow down the video

henpe071:
Updating the frame on the video by drawing the slider or clicking next frame is no longer tied to the speed slider's value.

Added notifys when the analysis slider value is changed to increase the speed and added some locks in the videowidget to prevent som sync problems.

Fixes #151

### DIFF
--- a/ViAn/Video/videoplayer.cpp
+++ b/ViAn/Video/videoplayer.cpp
@@ -114,6 +114,7 @@ void VideoPlayer::check_events() {
                 wait_load_read();
             } else if (current_frame != m_frame->load() && m_video_loaded->load()) {
                 set_frame();
+                qDebug() << "after set frame";
             }
         } else {
             // Timer condition triggered. Update playback speed if nessecary and read new frame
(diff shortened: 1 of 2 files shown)


niklasniklas:
Works fine!

ViAn/Video/videoplayer.cpp
```diff
@@ -115,13 +116,13 @@ void VideoPlayer::check_events() {
             } else if (current_frame != m_frame->load() && m_video_loaded->load()) {
                 set_frame();
             }
-        } else {
+
             // Timer condition triggered. Update playback speed if nessecary and read new frame
```
fredrikiselius: The comment is misleading as we no longer read the frame after a speed update

fredrikiselius (commented):
Codewise the changes looks fine. Haven't tried them yet. As a reminder, we need to be careful with locks in the main thread since if a lock is busy it will halt the gui until it the lock be acquired

fredrikiselius:
I've tested it now as well. Everything seems to be working fine

PR #155, opened 2018-10-08
State on video

henpe071:
Gave the video its own state which will update directly when the sliders in the manipulatorwidget are changed.

The apply button will update the current selected tag_frame or all tag_frames in the current label depending on what is selected in the project tree. It will also be disabled when the values haven't changed.

### DIFF
--- a/ViAn/GUI/mainwindow.cpp
+++ b/ViAn/GUI/mainwindow.cpp
@@ -119,7 +119,7 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent){
     manipulator_dock->close();
 
     connect(manipulator_wgt, &ManipulatorWidget::values, video_wgt, &VideoWidget::update_brightness_contrast);
-    connect(manipulator_wgt, &ManipulatorWidget::update_tag, video_wgt, &VideoWidget::update_tag);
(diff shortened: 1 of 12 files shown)


niklasniklas:
Bugs:
Didn't get the Apply for a label to work

Other comments (feature changes):
* The video and all of its' analysies should share the same state.
* All tags (both ordinalry and drawingtags) should have the same behaviour.
* If you are on a label or the "Drawing tag" and change brightness and press apply - all tags under it should be updated with the new brightness. Do not update zoom. 
* Every single tag (or drawingtag) should have a "update tag" with similar behaviour when re-tagging a tag using "T". This also updates zoom.

niklasniklas:
* Retagging av tag ("T") should also update brightness and contrast

* The label should not have its own brightness and contrast. 
** You want to be able to stand on a tag and adjust it (to a good b&c setting)
** then select the label and press apply for it to be applied to all tags
* If there is no "temporary" setting for the label to read, use the parent-video settings

henpe071:
Think I have fixed everything as wanted. However, as a side effect of fixing the update tag in the context menu, you can now use the arrow keys to navigate throught the project tree.

PR #162, opened 2018-10-15
Rotation

### LINKED ISSUE #116: Rotation and analysis
Make it possible to perform an analysis on a rotated video

henpe071:
Things rotates as  they should. This involves zoom rect, drawings and their bounding boxes, analysis selection rect, detection boxes and analysis details. They should all rotates with the video then it's rotated and handle drawing them on a rotated video.

NOTE:
Text does not rotate as everything else. Since text only consists of a single point and will always print horizontally you can't rotate it in the same way. Currently the the bounding box for the drawing rotates and the drawing will be printed from the point. I tried a quick solution to print the text on another mat and then rotate and fuse them back together. I didn't get it to work completly correct and if it would work the text would look like a watermark. This needs more discussion on how it should be but I save it for later in the meantime and put it as an issue.

Closes #116


niklasniklas:
Works fine!

ViAn/GUI/framewidget.cpp
```diff
@@ -189,7 +189,33 @@ void FrameWidget::set_anchor(QPoint p) {
...
-    if (m_vid_proj) m_vid_proj->get_video()->state.scale_factor = scale_factor;
+    if (m_vid_proj) {
+        m_vid_proj->get_video()->state.scale_factor = scale_factor;
+    }
+}
+
+/**
+ * @brief FrameWidget::set_rotation
+ * Slot function for updating framewidget after rotation have been done
+ * This will update the zoom rect coordinates
```
fredrikiselius: Should be mentioned that it updates the current video project state as well
```suggestion
 * Updates the zoom rectangle coordinates as well as the rotation held by the current video project state
```

ViAn/GUI/framewidget.cpp
```diff
@@ -430,8 +479,9 @@ void FrameWidget::mousePressEvent(QMouseEvent *event) {
         init_panning(event->pos());
         break;
     default:
+        // Rotate here
         bool right_click = (event->button() == Qt::RightButton);
-        emit mouse_pressed(scaled_pos, right_click);
+        emit mouse_pressed(rotate(scaled_pos, 360-m_rotation, true), right_click);
```
fredrikiselius: 360 should be made into a constant

ViAn/GUI/framewidget.cpp
```diff
@@ -455,7 +505,7 @@ void FrameWidget::mouseReleaseEvent(QMouseEvent *event) {
         end_panning();
         break;
     default:
-        emit mouse_released(scaled_pos, false);
+        emit mouse_released(rotate(scaled_pos, 360-m_rotation, true), false);
```
fredrikiselius: constant here as well

ViAn/GUI/framewidget.cpp
```diff
@@ -612,12 +661,28 @@ void FrameWidget::panning(QPoint pos) {
...
+ * @param pos           : Point to be rotated
+ * @param rotation      : Rotation
+ * @param swap          : If the width and height might need to be swaped
+ * @return
+ */
+QPoint FrameWidget::rotate(QPoint pos, int rotation, bool swap) {
+    int width = m_org_image.cols;
+    int height = m_org_image.rows;
+
+    if (swap && (rotation == 90 || rotation == 270)) {
```
fredrikiselius: fix constants

ViAn/Video/shapes/pen.cpp
```diff
@@ -45,14 +45,16 @@ cv::Mat Pen::draw(cv::Mat &frame) {
  * @param scale_factor - Zoom factor, used to scale drawing.
  * @return Returns the frame with drawing.
  */
-cv::Mat Pen::draw_scaled(cv::Mat &frame, cv::Point anchor, double scale_factor) {
+cv::Mat Pen::draw_scaled(cv::Mat &frame, cv::Point anchor, double scale_factor, int angle, int width, int height) {
```
fredrikiselius, line 54: Since this combination of parameters are used in several functions it might be a good idea to wrap them into a struct in order to increase readability

ViAn/utility.cpp
```diff
@@ -152,3 +152,34 @@ cv::Rect Utility::from_qrect(QRect rect) {
...
+ * @brief Utility::rotate
+ * Rotate the given point by the given rotation
+ * @param pos           : Point to be rotated
+ * @param rotation      : Wanted rotation
+ * @param width         : Width of frame
+ * @param height        : Height of frame
+ * @return              : Rotated point
+ */
+QPoint Utility::rotate(QPoint pos, int rotation, int width, int height) {
+    const double DEGREE_TO_RADIAN_FACTOR = M_PI / 180;
```
fredrikiselius: Should be assigned in the Utility namespace instead of the function so it isn't reassigned at every call

PR #178, opened 2018-10-24
Fix problems with the analysis play btn

### LINKED ISSUE #177: Problems with POI
* When POI-"Play" is selected it's only possible to use "Previous POI" a couple of times before it stops working. Issue when using left-right key and POI-selected. Seems to be a problem going from one selected (yellow) area to another. Also issues when navigating by double-clicking on the timeline.

* Sometimes POI-"Play" doesn't play the last Analysis area

henpe071:
* Fix some problems with the slider handle not jumping as intended when clicking the next/prev poi buttons. Now jumps to closes start even including the current poi. 
* Fix next/prev frame while analysis play is checked.  Will only jump to next frame that is a poi.
* Made the slider update when the analysisplay button is toggled. This will move the slider handle to the closest frame in a poi.
* The same will happen releasing the slider handle after having dragged the slider. The closes frame that is in a poi will be selected.
* Added some more lock guards to the affected atomic variables.

Fixes #177


niklasniklas:
Works fine! :-)

niklasniklas:
Issue: Playing with POI selected continues after last POI.

niklasniklas:
It stops att last POI when playing. However the POI button untoggles when reaching last POI. Also if you drag slider to the right of last POI POI button untoggles.

niklasniklas:
Small issue:
When at last POI and the program pauses, if you press play (or next frame) the video player advances one (1) frame each time. Otherwise fine!

niklasniklas:
ok to pull and create new issue with:
* press play, 
* next/prev
* pressing slider

PR #183, opened 2018-10-30
Updated all connects to the new form

henpe071:
All connects are now updated to the new version of using pointers instead of macros. This makes the code cleaner and easier to follow.  It also supposedly make it faster as well. Though the most important reason is that it's consistent.


niklasniklas:
No issues found!

PR #185, opened 2018-11-08
Interval

henpe071:
Added back intervals. Works like tags except you save two frames as an interval. These intervals can be merged, saved, loaded and other thing possible with tags.


niklasniklas:
Issues:
* Short "K" does not work
** Fixed
* Problem with next POI och prev POI button (todo?)
** Still some issues. Having problem when POI-play button is pressed. Merge with analysis-poi -branch?
* When tagging from interval, the Tag should not be automatically selected. When working with intervals you want to use the POI toggle button and able to tag from intervals. Should be the same for analysis. New tag appears in the tag witch has the checkbox checked. 
** works fine
* Bug: If you have the checkbox for a tag in video A checked and select an interval in video B, and slect new tag. THe tag appears in video A
** works fine
* Update New tag dialog with info for "Add new tag or check existing tag in project tree".
* Deleting last interval causes crash.
** Still crashes
* Merging intervals works fine! :-)

henpe071:
Added the last missing functions for the intervals to make them work with the analysis play button.

Will fix the delete crash tomorrow.

niklasniklas:
Strange issue:
* On newly loaded project: POI-play does not continue from one interval to next. 
* After I played an analysis and POI-played it the interval POI-play works
* However, the POI play then does not stop at last intervals end.

niklasniklas:
POI play works fine!

niklasniklas:
Hmmm... Crasch on line 157 in analysisslider.cpp when creating the first interval

niklasniklas:
Small issue
Does not POI-play if you do this sequence:
* start program
* new project
* open videos
* select one and create intervals
* toggle POI-play
* play from beginning
* the video stops at first POI-end

However if you select an other video and go back to the previous video the POI-play works...

niklasniklas:
Works fine!

PR #186, opened 2018-11-09
Image sequence updated

### LINKED ISSUE #105: Image sequence update
Already in progress
Removing images can be undone if project isnot saved

fredrikiselius:
The image sequences are now completely changed to use file hashes in the background.
Basically the new user functionality added is the ability to remove images from a sequence.

Closes #105 

**Issues:** _(16/11)_
- [ ] Deletion is not permanent on temporary projects
- [x] Projects item after the first image sequence are not displayed correctly (Merged with other branch?)
- [ ] Tree item image count not updated when an image is removed
- [x] Some images are skipped during the import process (_not!_ randomly)
_The issue arises when loading the tiff sequence. The reason is because many of the images are identical and as such, they have the same checksum. This is a problem since the file checksum is used to identify the images._
- [x] Slider incorrect after removing images
_Probably due to the issue of file skipping on import_


niklasniklas:
2 issues:
* The number i [ ] does not update until reloading project
* If the project is not saved with the sequence existing the deleting of a image in sequence does not create a temporary file. Not until you have saved with the seqence existing is the temporary file created.

However I would like to merge this as is and solve this in new branch. I'll discuss it with Henrik...

ViAn/Project/imagesequence.cpp
```diff
@@ -1,17 +1,36 @@
...
+ *
+ * This function is meant to be called before saving to file
+ */
+void ImageSequence::update() {
+    // If a hash from the saved order no longer exists in the unsaved order the file should be removed.
+    // If this happens the file should be saved as its hash instead of index
+    for (auto it = m_saved_order.begin(); it != m_saved_order.end(); ++it) {
+        if (m_unsaved_order.find((*it).first) == m_unsaved_order.end()){
+            if (!remove_image_from_disc((*it).first))
+                qDebug() << "failed to remove";
```
henpe071, line 20: Shouldn't it do something more when it fails to remove an item. I feel like it should at least print out what it failed to remove to help the user identify the problem.

ViAn/Project/imagesequence.cpp
```diff
@@ -20,95 +39,262 @@ ImageSequence::ImageSequence(const std::string& path) : Video(true){
...
  * @brief ImageSequence::get_pattern_name
  * Returns the string neccessary for opencv's VideoCapture class to open and play the sequence.
- * The string is a path relative the project path and need as such be appended to that before playback.
+ * The string is a path relative the project path.
  * @return sequence pattern
  */
 std::string ImageSequence::get_pattern_name() {
-    int digits = Utility::number_of_digits(length());
-    return "%0" + std::to_string(digits) + "d";
+//    int digits = INDEX_LENGTH;//Utility::number_of_digits(length());
```
henpe071, line 72: Remove comment?

ViAn/imageimporter.cpp
```diff
@@ -26,16 +20,21 @@ ImageImporter::ImageImporter(const QStringList& images, const QString& dest, QOb
  */
 void ImageImporter::import_images() {
     int num_images = m_images.size();
-    int num_digits = Utility::number_of_digits(num_images);
+    int num_digits = 10;//Utility::number_of_digits(num_images);
```
henpe071, line 24: remove comment & add as constant

ViAn/utility.cpp
```diff
@@ -155,6 +165,47 @@ QRect Utility::from_cvrect(cv::Rect rect) {
     return QRect(from_cvpoint(rect.tl()), from_cvpoint(rect.br()));
 }
 
+QByteArray Utility::checksum(const QString &file, QCryptographicHash::Algorithm hash_algorithm) {
```
henpe071, line 171: Add a comment

ViAn/utility.cpp
```diff
@@ -155,6 +165,47 @@ QRect Utility::from_cvrect(cv::Rect rect) {
...
+    while (it.hasNext()) {
+        auto path = it.next();
+        auto current_hash = checksum(path);
+        for (auto hash : hashes) {
+            if (current_hash == hash) {
+                QFile(path).remove();
+                break;
+            }
+        }
+//        if (checksum(path) == hash) {
```
henpe071, line 203: Remove comment

henpe071 (commented):
Added some comments. 

But I hade some troubles loading sequences. I selected 21 images to import (0-20) from a folder but only 10 of the file were actually imported to the program. And it wasn't even like the 10 first or last. It was more random. Got number 2,5,7,10... I haven't tried if this problem exists in other branches yet but it's not something i have noticed.

henpe071:
Deleting an image in a sequence before saving the project will not save the deletion when saving and the removed image will be back later after the project is restarted.

Also, the slider is not correct. It works before saving and closing. I have 8 images (2,5,7,12,14,17,19,20). Had 10, nr 10 and 16 (i think) were removed before saving and closing. The slider max is now shown as 9 and then will try to jump to the frame given by the image in the sequence. For example clicking the image with number 17 will jump to frame 15 which is more than the 9 the slider says is max. 17 will try to jump to 15 because 2 frames before it were removed 17-2=15 so this is also a problem.

Then the problem Niklas found and wrote about on #191 .

>  When loading a project with more than one image sequence: All object after the first image sequence in project tree is not loaded properly. They appear as object named "placeholder". When selecting placeholder the porgram crasches

That the program crashes when clicking placeholder should probably get a fix so it just get a warning or similar, can make a new issue, but the problem here is that you get the placeholders.

ViAn/imageimporter.cpp
henpe071: also, num_images is unused

fredrikiselius:
**Issues:**
- [ ] Deletion are not permanent on temporary projects
- [ ] Slider incorrect after removing images
- [ ] Projects item after the first image sequence are not displayed correctly (Merged with other branch?)
- [ ] Tree item image count not updated when an image is removed
- [ ] Some images are skipped during the import process (randomly)

niklasniklas:
It appears to have probelm loading loading two identical images. I loaded 15 images. And for ex. img 2 and 3 were identical and only 3 is shown in project tree. But I received 15 on slider

PR #191, opened 2018-11-15
Delete video crash fixed

### LINKED ISSUE #190: Deleting videos wrong way cause crash
When selecting a list of videos down to up and then deleting, the program will crash.

Deleting the list the wrong way.

henpe071:
Should hopefully fix the crash when deleting all videos while selecting them from the bottom up

Fixes #190

### DIFF
--- a/ViAn/GUI/TreeItems/videoitem.cpp
+++ b/ViAn/GUI/TreeItems/videoitem.cpp
@@ -9,6 +9,7 @@
 #include "opencv2/core/core.hpp"
 #include "opencv2/videoio/videoio.hpp"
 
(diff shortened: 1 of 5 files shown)


niklasniklas:
Works fine with video. :-)

### Tested the below on "checksum" branch and its works ok. Test the below when this branch and checksum are merged.
Some issues with image sequence. (Both deleting and saving/opening)
* Crasches when removing image sequence
* When loading a project with more than one image sequence: All object after the first image sequence in project tree is not loaded properly. They appear as object named "placeholder". When selecting placeholder the porgram crasches.

PR #196, opened 2018-11-16
Fix crash when clicking placeholder

henpe071:
Can be hard to test but this is a fix to prevent a crash after something has gone wrong. Since when the item has the text placeholder it doesn't have a videoproject stored, I just return when that happend.


henpe071:
Might fix for the other items later

niklasniklas (approved):

PR #197, opened 2018-11-19
Fixed some cases where slider went outside the interval while ana btn

### LINKED ISSUE #189: Slider ana btn bugs
The slider can sometimes get outside the poi when the analysis button is checked.
This can happen when:

pressing play,
pressing next/prev frame
pressing slider
change item in the tree

henpe071:
Fixes some cases where the slider would go outside the interval while the analysis play button was checked.

Fixed pressing next/prev frame, playbutton and clicking on slider.

Also removed some unused variables.

Fixes #189


niklasniklas:
Works fine! :-)

PR #198, opened 2018-11-21
New settings

henpe071:
Addad a settings dialog under options in the edit menu.

Here the user can customize some settings which are save in a singleton.
Alot of other settings are also store in the same file and accessed through the Constants namespace


niklasniklas:
Works fine! :-)

PR #199, opened 2018-11-23
stdstring -> qstring

henpe071:
Since the program is using mostly qcomponents, using qstring will make things easier and reduce the amound of conversions needed.

on most places, not in the analysisfiles


niklasniklas:
works fine

PR #200, opened 2018-11-23
Seq wrong name

henpe071:
Exported frame and the image in the report will now have the original image's name.
TODO Does not give the original name if exported from the Frameexporterdialog in the export menu.

The tooltip on videoitems in the project tree and the path in the frameexporter dialog does no longer include the "%010d" string.

When opening the frameexporter dialog while no interval is selected the current frame is chosen as a default.

Also added some default constants in analysisslider for when no interval is selected.

Continuation of #186 . Merge that one first


niklasniklas:
Take away the extra ".tiff" from that is printed in the report. Only reference to the original filename and not to the new exported filename. It is still ok to export it with the extra ".tiff" to the Stills directory.

niklasniklas:
remove ".tiff" and keep the original suffix

niklasniklas:
works fine

PR #201, opened 2018-11-28
Copy the current frame to the clipboard

henpe071:
Added an option in the export (?) menu to copy the current frame to the clip board to use in other applications.

Also set a shortcut, Shift, alt, c (?).


niklasniklas:
Ctrl+C with zoomtool selected causes crasch...sometimes (if pressed multiple times). Have not found any systems for the crasch
Maybe we sholud override this to copy current frame to clipboard afterall.

niklasniklas:
Works fine!

PR #215, opened 2018-12-17
Different seq with checksum

henpe071:
Added an update_frame_size function and a skip_update variable. These should fix (avoid) the problems.
I also moved the set_zoom_tool call to the on_video_info function to prevent the extra call to the processor. Might be able to fix these in a better way since it now only sets the tool when changing the video and not when changing between anything in the tree. But for now I'm just happy it work so I'll leave it at that. 

This branch is not yet completely finished as i still have to save the state for the Sequencetagitems and clean it, but put it up so it can be tested and commented on.


niklasniklas:
Having some issues with zoom-rect moving between tags. Problem with anchor?
Tested video-tags and they appear to have received a similar issue, which it does not have in the "bookmark" branch.

Buttons for tagging and label should be disabled. Untagging "U" causes crash.

PR #217, opened 2018-12-18
New diff seq

henpe071:
The new solution for the tag sequences. Saves them as normal tags. And saves the state in the tag.

Not completely done yet but putting it up here so it can be tested.
Known problems right now: 
 - The first image clicked after loading will decide the zoom lvl. Something with fit viewport have to be looked at.
 - There seems to be a crash when closing the project.

------------

And yes i merged the bookmark order branch. Since i wanted the latest code in case that branch fixes some issues.

----------------------------
Newset edit:
It's mostly done, at least from the user's side.
The sync problem should be fixed or at least be a lot better with the new "frame_changing" variable from the videowidget. This variable tells the processor that there will come a new frame so it doesn't update the old frame with the new frame's state.
I sped through the changed towards the end, because I wanted to be as finished as possible before the holidays. This means that the deletions of single frames and of the whole tag on a tag sequence is not done correctly. But it should work from the users side (not crash). It might leaves some memory left uncleaned and some files left in the folder (in the case of the deleting a single frame).

### DIFF
--- a/ViAn/Analysis/analysissettings.cpp
+++ b/ViAn/Analysis/analysissettings.cpp
@@ -2,7 +2,7 @@
 
 #include "constants.h"
 
-AnalysisSettings::AnalysisSettings(ANALYSIS_TYPE type,
+AnalysisSettings::AnalysisSettings(int type,
                                    std::pair<int, int> interval,
                                    cv::Rect bounding_box)
     : type(type)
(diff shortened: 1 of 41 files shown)


niklasniklas:
Issue:
Having som issues with loading saved imq sec. Craches.
After saving the images that are imported as tags are not the images name but numbers.
Having some issues with re-zooming to fit-viewport with the first image in image seq.

niklasniklas:
Tested some and seem to be working ok.

PR #218, opened 2018-12-20
Videostate are saved seperately

henpe071:
Made the videostate its own writeable to make the savefile easier to read and it's less repetetive code when wanting to save a videostate.

Edit: This code is included in the #217 pull request


niklasniklas:
ok to pull

PR #220, opened 2019-01-25
Delete frames in tag sequence

henpe071:
Frames are now removed temporarily and the name is changed to the hash. It's removed completly when the project is saved, otherwise it's restored. If the sequence isn't saved from the beginning the frame is just removed.

Continuation of #217


niklasniklas:
Almost there.
Issue: 
If I do the following twice:
* delete an image
* reopen the same vian-file and select "no" if want to save

the program crasches.

PR #221, opened 2019-01-28
Bad name check

henpe071:
Added a check when saving the project. Now only accepts upper and lower case letters and numbers. This to prevent a crash when entering åäö and special character.

Also a continuation of #217


niklasniklas:
Need to expand allowed charcters with 
" " 
"_" and 
"-"

PR #222, opened 2019-01-29
Text rotation

### LINKED ISSUE #163: Rotate text
Find a way to rotate the text correctly

henpe071:
The bounding box now rotates with the text drawing.

Fixes #163


niklasniklas:
Seems that the coordinates when doubleclicking is wrong, when canvas is rotated or zoomed.
I managed to select an object by doubleclicking outside the object (somewhere else in the viewport)

The coordinates when singleclicking is ok. If you select an object by clicking it in in Drawings-widget, then single click works.

niklasniklas:
Otherwise ok!

niklasniklas:
Works fine! :-)

PR #223, opened 2019-02-01
Minimized the sync problem

henpe071:
Moved some signals around to minimize the sync problems. Also, clicking a drawing from the drawing widget now sets the frame to fit screen.

Remove the state as a pointer as it made some problems to the normal tags and when the state was sent around. The state a tagframe was holding was updated as it was sent around. What exactly was sent wasn't certain in all places.
Removed the pointers and the tags seems to work again and the tag sequence deletion that the pointers were supposed to fix also seems to work now. This might need some more through testing with all tags variations and deletions as this was fixed quickly just before the clock became too much.

### DIFF
--- a/ViAn/GUI/DrawingItems/frameitem.cpp
+++ b/ViAn/GUI/DrawingItems/frameitem.cpp
@@ -18,3 +18,8 @@ Shapes* FrameItem::get_shape() {
 int FrameItem::get_frame() {
     return m_frame;
 }
+
+void FrameItem::set_frame(int frame) {
+    m_frame = frame;
+    setText(0, QString::number(frame));
+}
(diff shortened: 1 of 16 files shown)


niklasniklas:
Panning in image-seq-tag is not saved. Always centre
Tags in video does not work i development...

niklasniklas:
- [x] Pressing the label "Sequence" for a image tag seq changes the first tag state to fit-to-screen

- [x] Issue whne deleting an image in an image seq. The drawings in the "Drawings" widget does not get updated which leads to strange behavior

- [x] Shortcut "T" still works even if image-tag-seq

- [ ] Genrated report closes when Vian is closed (Save for later?)

- [x] Thumbnails in project window seems to be mixed RGB to BGR? They are saved in _thumbnail-folder?

- [x] Tag drawings doesnt seem to be working

niklasniklas:
- [x] When an analysis is done and you navigate to a POI. THen clicking the main video (parent to analysis) it jumps to a different frame

PR #224, opened 2019-02-04
Connect Bookmark widget and projectwidget

henpe071:
Selecting a bookmark in the bookmarkwidget will now select in the projectwidget.

This branch builds on #223 but it's not sure that it needs that functionality (fixes)


niklasniklas:
* Issue when: Double clicking a bookmark when the right video is already selected. Video does not jump to correct frame. If you create two bookmarks from same video, you cant jump to them

niklasniklas:
Works fine

PR #225, opened 2019-02-06
Update resource file with the currently used icons

henpe071:
Resource now points to the new ones, and the old ones are removed.


niklasniklas:
works fine after clean

PR #228, opened 2019-02-07
Loop

henpe071:
Added a loop function that will loop over the current interval


niklasniklas:
Works fine.
Question: Is it possible to save the selection for each video when jumping between videos?

niklasniklas:
Works fine. 

However found an issue on this branch but maybe its also on other pull req. 
* When opening a tag-seq. Alot of buttons are disabled.
* When selecting a video, the buttons are not enabled.

henpe071:
I recall fixing that somewhere. I can look into it if it doesn't get fixed once everything is merged.

PR #232, opened 2019-03-05
Flip

henpe071:
Added horizontal and vertical flip.

Shortcuts H and V


niklasniklas:
Have quite a few issues :
* Bookmarks does not get the right state.
* Tags does not get the right state
* Tags does not update when pressing "T"

niklasniklas:
Maybe we should define H and V to apply on the original and not current?

niklasniklas:
Works fine.
Found an other issue:
* I created a img seq and saved proj
* I created an other img seq and deleted it
* when saving the program crashes

PR #235, opened 2019-03-25
Fix seq crash

### LINKED ISSUE #234: Crash when img seq deleted
Found an other issue:

I created a img seq and saved proj
I created an other img seq and deleted it
when saving the program crashes

henpe071:
The sequence was not removed from the the list of vid_projs in the project

Should fix #234


niklasniklas:
Seems to be working fine!

PR #237, opened 2019-05-27
Ana widget + yolo

henpe071:
Added yolo analysis as object detection and a new widget for a better overlook of the analysis.

Updated the rest of the program to work with two different kinds of analysis and made some small QoL changes for clarity.

Removed some old files that are no longer needed.


niklasniklas:
When saving an object detection, only the start of each interval is saved
Test case: 

- Run an object detection
- If you have multiple detections in a row you get a solid yellow marking on timeline, but in Yolo widget you get all frames from the detections in dropdown list
- when project is saved and reloaded only the start of each "yellow interval" is loaded i yolo-dropdown

niklasniklas:
Crasches when opening "Object detection settings." and pressing OK.

henpe071:
Pushed up what I've made currently if you want to try it out.
Note: Motion detection will currently crash the object detection widget if tried to open.

niklasniklas:
Settings changed in Object detection settings does not get "saved". If I change settings the settings are changed but when I open the settingsdialog the default values are shown. THis only applies to network size or sample freq

niklasniklas:
"End frame numer" (to the right of the timeslider at the bottom) i Yolowidget is not always initialized properly when opening Yolowidget

niklasniklas:
Two thigs for Object detecton widget:
* A checkbox for automatically update videoplayer if checked (and not update if unchecked). Default unchecked
* It is enough with "heltal" for confidence under each thumbnail

niklasniklas:
Change default: Menue: Video analysis->Analysis input should be checked by default.

niklasniklas:
Sometimes, when I change sample freq for object detection (for example 0.2 or 0.1 when I have 29 fps on video) every frame is analyzied. The rounding to of the specified sample frequency does not seem to work all the time...

niklasniklas:
Great improvements!

PR #238, opened 2019-06-18
Fixed some merge problems and other small bugs

henpe071:
Cleaned the yoloanalysis, adjusted the splashscreen duration, new splashscreen and removed some debugs


niklasniklas:
2 crasches found:
* Doublecklick on analysis that is not finished (Loading... xx%)
* Opening an analysis with different threshold than that one that is previously opened in "Yolowidget".

niklasniklas:
In
`int main(int argc, char *argv[])`

Replace:
`Sleep(2000);`
`splash.close();`

with
` QTimer::singleShot(1000, &splash,SLOT(close())); // Timer`

Does not solve the "Exit ViAn" problem

niklasniklas:
Pushed some fixes.
* New vian-help.pdf
* Timer for splash
* New name for report-documents

henpe071:
I pushed up what I made so far.

I added so video sequences have thumbnails. However, thumbnails take a while to generate so the program will feel slow on startup or when adding a new sequence. Generating these thumbnails need to be done on a seperate thread, like how the yolo widget generates the thumbnails. 

I will wait with the thumbnails on the tag sequences because they would get the same problem. Currently it will create new thumbnails everytime. Fixing that will speed it up when loading the project but not when adding a new sequence. 

You can try this out and I will see if i have time to fix everything tomorrow.

